PR #118, opened 2014-03-22
Add several type-safe property retrieval methods

MariusVolkhart:
This change makes the document class much easier to use, and the
resulting code much easier to read since a lot of manual type casts
are now avoided. It also enforced type safety in the code, and puts
the onus on the developer to make it clear what the type
expectations are.

### DIFF
--- a/src/main/java/com/couchbase/lite/Document.java
+++ b/src/main/java/com/couchbase/lite/Document.java
@@ -231,6 +231,61 @@ public final class Document {
     }
 
     /**
+     * Shorthand for (String) getProperties().get(key)
+     */
+    @InterfaceAudience.Public
+    public String getStringProperty(String key) {
+        return (String) getProperty(key);
+    }
+
+    /**
+     * Casts the requested property, using the default if null is encountered
+     */
+    @InterfaceAudience.Public
+    public double getDoubleProperty(String key, double defaultValue) {
+        Object o = getProperty(key);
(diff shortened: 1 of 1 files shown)


sergio91pt:
I share the same opinion as snej, IO cannot be trusted. Another app may be using the same database and for example emits doubles instead of int's or the user pokes in the database and bad data gets replicated. Either way we should gracefully handle the situation, specially when dealing with more than 1 document, like in a Map Reduce function.

This is what I use:

```Java
static <T> T getProperty(Map<String, ?> map, String key, Class<T> clazz) {
    Object value = map.get(key);
    if (value == null) {
        return null;
    } else if (clazz.isAssignableFrom(value.getClass())) {
        return (T) value;
    }
    String actual = value.getClass().getName();
    String expected = clazz.getName();
    Object docId = map.get("_id");
    Log.w(TAG, "Invalid value in key " + key + " of " + docId
            + ". Expected " + expected + " but got " + actual);
    return null;
}

static Integer getInteger(Map<String, ?> map, String key) {
    Number value = getProperty(map, key, Number.class);
    if (value == null) {
        return null;
    } else if (value instanceof Integer) {
        return (Integer) value;
    } else {
        return value.intValue();
    }
}
```

hideki:
@MariusVolkhart,
In short term, we will not apply this change. So I will close this PR.
Thank you very much for your contribution!

PR #228, opened 2014-06-12
Prevent modification to nested property maps

mjq:
For consideration. I don't know if this is actually the route CBL should go in, but it's a possibility. This does introduce some map copying in `getProperties`, making it more expensive, but only in the case when there are nested maps.

See couchbase/couchbase-lite-java-core#226.

### DIFF
--- a/src/main/java/com/couchbase/lite/SavedRevision.java
+++ b/src/main/java/com/couchbase/lite/SavedRevision.java
@@ -24,6 +24,7 @@ import com.couchbase.lite.util.Log;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/SavedRevision.java
```diff
@@ -201,7 +202,33 @@ public long getSequence() {
...
+
+        for (Map.Entry<? extends K,? extends V> entry : m.entrySet()) {
+            if (entry.getValue() instanceof Map) {
+                if (copy == null) {
+                    copy = new HashMap<K,V>(m);
+                }
+                Map newValue = nestedUnmodifiableMap((Map) entry.getValue());
+                try {
+                    copy.put(entry.getKey(), (V) newValue);
+                } catch (ClassCastException e) {
```
sergio91pt: Due to type erasure of Java Generics, this try catch is not needed and won't prevent an Object that is not assignable from V to be added to the new map.

hideki:
@mjq,
In short term, we will not apply this change. So I will close this PR.
Thank you very much for your contribution!

PR #363, opened 2014-12-20
Fixed #299 Improve Retry behavior

### LINKED ISSUE #299: Improve retry behavior when failing to push a doc
During dogfood testing, I noticed that in a continuous push replication, if it failed to push a doc after a few tries, it would give up.  The only thing that would trigger it to retry again was a network change notifications.

OTOH, on iOS it would continue to retry.

Test which reproduces this issue:

https://github.com/couchbase/couchbase-lite-android/blob/712095f3e13d4fc7d2d1fa29f85ae4c53e5b6d57/src/androidTest/java/com/couchbase/lite/replicator/ReplicationTest.java#L847

hideki:
1. Retry new replication session if previous session ends with error see ReplicationInternal.java
2. Apply same retry logic in a replication session. see RemoteRequestRetry.java

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -423,7 +423,7 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
                         Map<String, Object> res = (Map<String, Object>) result;
 
                         if (e != null) {
-                            setError(e);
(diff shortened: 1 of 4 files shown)


tleyden:
> Adding ACTIVE, and isolate IDLE to OK and ERROR are great step. Adding new states and trigger could effect other part of codes. Can we track this by new issue ticket?

I just added: https://github.com/couchbase/couchbase-lite-java-core/issues/369

PR #365, opened 2014-12-23
Feature/issue 352 get session and checkpoint when retrying replication

### LINKED ISSUE #352: Get session and checkpoint when retrying replications
From @snej:

I found and fixed some Obj-C replicator bugs today. A few of them might also appear in the Java (and C#) code; it’d be worth taking a look.

couchbase/couchbase-lite-ios@3ff9c7a — When retrying a replication, make sure to get session & checkpoint. Especially important if the replication initially failed to create a session or fetch the remote checkpoint, because then we don't have one at all. (This could have been causing checkpoint conflict errors.)

(And I forgot to note this in the commit log, but this patch also makes the replicator stop, even if it’s continuous, when it receives a permanent-type error. The bug I observed was that when auth fails, the 401 status from getting the remote checkpoint wasn’t stopping the replicator. Instead it would keep trying once a minute.)

Note: I factored this ticket out of https://github.com/couchbase/couchbase-lite-java-core/issues/265

hideki:
NOTE: This PR depends on #363 (issue number #299)

Fix #352 Get session and checkpoint when retrying replications

- This fix depends on the fix of #299
- Case 1 - By fixing #299, case 1 is also addressed.
- Case 2 - iOS version: stop replicator immediately when setError() was called with permanent error. But, for Core Java version, stop replicator when state becomes IDLE.
              -  Two test cases which calls /{db}/_all_docs are broken if stop replicator for HTTP error 406, because test case does not implement /{db}/_all_docs. Because of this, current version eliminate 406 http error from permanent error list.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -423,7 +423,7 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
                         Map<String, Object> res = (Map<String, Object>) result;
 
                         if (e != null) {
-                            setError(e);
+                            setError(e, "/_all_docs?include_docs=true");
                             revisionFailed();
 
                             // TODO: There is a known bug caused by the line below, which is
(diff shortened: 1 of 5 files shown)


tleyden:
Merged by hand.

tleyden:
I just cherry picked https://github.com/couchbase/couchbase-lite-java-core/commit/d329d4a668f2ebe4f563bc1d864d31c67d5a5467, since all of the other commits were already on master.

PR #366, opened 2014-12-23
Feature/issue 316 fail to resolve old conflicts

### LINKED ISSUE #316: New revisions to resolve old conflicts fail insert constraints
https://github.com/couchbase/couchbase-lite-ios/issues/509

hideki:
Fix #316  Get session and checkpoint when retrying replications

- Because reproduce steps are not clear now, so no test case for this fix.
- Basically ported fix for iOS - https://github.com/couchbase/couchbase-lite-ios/commit/7b3fca59e88b2281802425fa0c8d3659c8186a6a#diff-bb5cfc8dca22d97f47291aa05a9f5bebR350
  - String winningRevIDOfDoc(long docNumericId, AtomicBoolean outIsDeleted, AtomicBoolean outIsConflict) throws CouchbaseLiteException throws exception if encountered error. No change.
  - In public RevisionInternal putRevision(RevisionInternal oldRev, String prevRevId, boolean allowConflict, Status resultStatus) throws CouchbaseLiteException, if newSequence <= 0 (conflict), return after updating parentSequence.
  - Because of current AndroidSQLiteStorageEngine implementation, unable to get detailed error code of SQLite. we might need to work for it later.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -2391,6 +2391,12 @@ public final class Database {
     /**
      * Returns the rev ID of the 'winning' revision of this document, and whether it's deleted.
      * @exclude
+     *
+     * in CBLDatabase+Internal.m
+     * - (NSString*) winningRevIDOfDocNumericID: (SInt64)docNumericID
+     *                                isDeleted: (BOOL*)outIsDeleted
+     *                               isConflict: (BOOL*)outIsConflict // optional
+     *                                   status: (CBLStatus*)outStatus
      */
     @InterfaceAudience.Private
     String winningRevIDOfDoc(long docNumericId, AtomicBoolean outIsDeleted, AtomicBoolean outIsConflict) throws CouchbaseLiteException {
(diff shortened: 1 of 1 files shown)


tleyden:
> NOTE - In AndroidSQLiteStorageEngine.java, CBL Android uses insert() method of SQLiteDatabase
> which return -1 for error case. Instead of insert(), should use insertOrThrow method()
> which throws detailed error code. Need to check with SQLiteDatabase.CONFLICT_FAIL.

What advantage would switching to `insertOrThrow` give us?  

@pasin I'm assuming that couchbase-lite-java-native would require some work to implement `insertOrThrow`?

> However, returning after updating parentSequence might not cause any problem.

Can you explain this in more detail?

hideki:
>insertOrThrow()

In case of encountering error, insertOrThrow() method can throw Exception with detailed information. But insert() returns only -1. Logic could be different for each error case.

>Can you explain this in more detail?

CBL iOS return immediately if error code is CONFLICT_FAIL. otherwise, return after updating parentSequence. CBL core java can not handle CONFLICT_FAIL. So always return after updating parentSequence. However, this difference might not cause any issue. I am not 100% sure.

tleyden:
I see, makes sense.  Can you update the comments in the code to mention what you said in the above github comment?

pasin:
The java-native doesn't have insertOrThrow implemented so we will need to implement it. Not sure if it's simple as trying to get the last error code/message from the sqlite then throw the exception.

PR #370, opened 2015-01-08
Fix #304 Implement Query.sort

hideki:
- Java does not have completely equivalent implementation of NSSortDescriptor. Instead of NSSortDescriptor, we use java.util.Comparator.
- Comparator is primitive interface. So CBL Java Core can not provide high-level sorting descriptor like CBL iOS implementation.


pasin:
I think this issue is related to Product/API specification. @zgramana , can you review this?

PR #390, opened 2015-01-20
fix #172 support gzip

### LINKED ISSUE #172: Add gzip support

hideki:
Gzip support for replication #172 

- enable compression if sync gateway version is higher than 0.92 (first version SG support gzip)
- request body is gzipped if json size is > 100 (client -> SG)
- response body is gzipped if json size is > 1000 (SG -> client)  note: this is SG side configuration.
- binary data is not gzipped
- Test is done manually by using ToDo Lite app (Create text > 100 for request test, create text > 1000 for response test)
- Unit test for Android was also updated to handle gzipped request.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
+++ b/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
@@ -12,6 +12,7 @@ import com.couchbase.lite.support.RemoteRequest;
 import com.couchbase.lite.support.RemoteRequestCompletionBlock;
 import com.couchbase.lite.util.CollectionUtils;
 import com.couchbase.lite.util.Log;
+import com.couchbase.lite.util.Utils;
 
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -175,13 +176,11 @@ public class BulkDownloader extends RemoteRequest implements MultipartReaderDele
                         inputStream = entity.getContent();
 
                         // decompress if contentEncoding is gzip
-                        Header contentEncoding = entity.getContentEncoding();
-                        if(contentEncoding != null && contentEncoding.getValue().contains("gzip")){
+                        if(Utils.isGzip(entity)){
                             inputStream = new GZIPInputStream(inputStream);
                         }
 
                         try {
-
                             fullBody = Manager.getObjectMapper().readValue(inputStream,
                                     Object.class);
                             respondWithResult(fullBody, error, response);
--- a/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
+++ b/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.zip.GZIPInputStream;
 
 /**
  * A special type of RemoteRequest that knows how to call the _bulk_get endpoint.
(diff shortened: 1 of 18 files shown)


src/main/java/com/couchbase/lite/support/MultipartDocumentReader.java
```diff
@@ -37,26 +39,50 @@ public MultipartDocumentReader(HttpResponse response, Database database) {
...
     }
 
-    public void setContentType(String contentType) {
-        if (contentType.startsWith("multipart/")) {
+    public void setHeaders(Map<String, String> headers){
+        String contentType = headers.get("Content-Type");
+        if(contentType.startsWith("multipart/")){
+            // Multipart, so initialize the parser:
             multipartReader = new MultipartReader(contentType, this);
             attachmentsByName = new HashMap<String, BlobStoreWriter>();
```
tleyden: This line `attachmentsByName = new HashMap<String, BlobStoreWriter>();` seems to be duplicated

src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
```diff
@@ -170,8 +171,17 @@ protected void executeRequest(HttpClient httpClient, HttpUriRequest request) {
                 else {
                     Log.v(Log.TAG_SYNC, "contentTypeHeader is not multipart = %s",contentTypeHeader.getValue());
                     if (entity != null) {
+
+                        inputStream = entity.getContent();
+
+                        // decompress if contentEncoding is gzip
+                        Header contentEncoding = entity.getContentEncoding();
+                        if(contentEncoding != null && contentEncoding.getValue().contains("gzip")){
```
tleyden: Can you factor this:

```
if(contentEncoding != null && contentEncoding.getValue().contains("gzip")){
```

into an isGzip(contentEncodingHeader) method?

Also this code below can be changed to call that same method:

```
    protected void startJSONBufferWithHeaders(Map<String, String> headers){
        jsonBuffer = new ByteArrayBuffer(1024);
        String contentEncoding = headers.get("Content-Encoding");
        jsonCompressed = contentEncoding != null &&  contentEncoding.indexOf("gzip") >= 0;
    }
```

src/main/java/com/couchbase/lite/support/RemoteMultipartRequest.java
```diff
@@ -44,7 +44,7 @@ public void run() {
             throw new IllegalArgumentException("Invalid request method: " + method);
         }
 
-        request.addHeader("Accept", "*/*");
+        request.addHeader("Accept", "*/*"); // should be "application/json"??
```
tleyden: I think it's `*/*` because the server might return `multipart/related` instead of `application/json`

src/main/java/com/couchbase/lite/support/RemoteRequest.java
```diff
@@ -317,10 +300,66 @@ public void respondWithResult(final Object result, final Throwable error, final
...
+        }
+    }
+    /**
+     * gzip
+     *
+     * in CBLRemoteRequest.m
+     * - (BOOL) compressBody
+     */
+    protected ByteArrayEntity setCompressedBody(byte[] bodyBytes){
+        if(bodyBytes.length < 100){
```
tleyden: Can this use MIN_JSON_LENGTH_TO_COMPRESS instead of hardcoded 100?

Also, should it throw an exception or print a warning here instead of returning null?   If it returns null, does that mean that content is never sent?

src/main/java/com/couchbase/lite/support/RemoteRequest.java
```diff
@@ -317,10 +300,66 @@ public void respondWithResult(final Object result, final Throwable error, final
...
+            return null;
+        }
+
+        // Gzipping
+        byte[] encodedBytes = Utils.compressByGzip(bodyBytes);
+
+        if(encodedBytes == null || encodedBytes.length >= bodyBytes.length) {
+            return null;
```
tleyden: should it throw an exception here?

src/main/java/com/couchbase/lite/support/RemoteMultipartRequest.java
hideki: Thank you for your comment!

hideki:
@tleyden Please re-review updated codes?

PR #398, opened 2015-01-27
Fix #328

### LINKED ISSUE #328: Excessive checkpointing during push replication
Noticed while investigating https://github.com/couchbase/couchbase-lite-java-core/issues/313

To reproduce:

* Run [office-radar-android](https://github.com/tleyden/office-radar-android) from scratch
* Re-run app 
* During push, excessive number of "PUT checkpoint" requests (as compared to iOS)

hideki:
- To change the rule of sending PUT /{db}/_local/xxxx to follow iOS implementation
- Cleanup unused codes. By shifting to StateMachine, some variables and methods which are from iOS code were not used.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -561,9 +561,7 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
     }
 
     private void revisionFailed(RevisionInternal rev, Throwable throwable) {
-        if (Utils.isTransientError(throwable)) {
-            // retry later
-        } else {
+        if (!Utils.isTransientError(throwable)) {
             Log.v(Log.TAG_SYNC, "%s: giving up on %s: %s", this, rev, throwable);
             pendingSequences.removeSequence(rev.getSequence());
         }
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/replicator/PullerInternal.java
```diff
@@ -566,7 +562,7 @@ public int compare(RevisionInternal reva, RevisionInternal revb) {
 
     private void revisionFailed(RevisionInternal rev, Throwable throwable) {
         if (Utils.isTransientError(throwable)) {
-            revisionFailed();  // retry later
+            // retry later
```
pasin: Maybe we shouldn't keep this if block as it does nothing.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -1314,11 +1319,6 @@ protected void retryReplicationIfError(){
         }
     }
 
-    // TODO: not used, please remove
-    protected void updateActive() {
-        Log.v(Log.TAG_SYNC, "%s: updateActive() called", this);
-    }
-
```
pasin: As the updateAction() method is now removed, we should update some log messages in the file that say something about calling updateActive() method to avoid confusion.

PR #414, opened 2015-02-06
Fixed #413 NullPointerException in query with setGroupLevel() without re...

### LINKED ISSUE #413: NullPointerException in query with setGroupLevel() without reduce function and no data
Following #355, there's still a `NullPointerException` that slipped through!

Inside `View.reducedQuery()` ([currently on master](https://github.com/couchbase/couchbase-lite-java-core/blob/c59086252fa30c4cb77eca62f60577b61c88f1cb/src/main/java/com/couchbase/lite/View.java#L920)):
```java
List<Object> keysToReduce = null;
// ...
if(getReduce() != null) {
    keysToReduce = new ArrayList<Object>(REDUCE_BATCH_SIZE);
    // ...
}
// ...
cursor.moveToNext();
while (!cursor.isAfterLast()) {
    // ...
}
if(keysToReduce.size() > 0) {
```
As you can clearly see from the code above on the last line `keysToReduce` can be `null` when there's no reduce function and the query contains no data, so instead of `keysToReduce.size() > 0` it should be `keysToReduce != null && keysToReduce.size() > 0`.

hideki:
...duce function and no data

- Just added null check

### DIFF
--- a/src/main/java/com/couchbase/lite/View.java
+++ b/src/main/java/com/couchbase/lite/View.java
@@ -906,7 +906,7 @@ public final class View {
 
         }
 
-        if(keysToReduce.size() > 0) {
+        if(keysToReduce != null && keysToReduce.size() > 0) {
             // Finish the last group (or the entire list, if no grouping):
             Object key = group ? groupKey(lastKey, groupLevel) : null;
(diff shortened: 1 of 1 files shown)


hideki:
@pasin Could  you please review codes. This is minimal code change. I will also send PR for unit test on Android.

PR #418, opened 2015-02-09
Fixed #242 - Nested transaction issue

### LINKED ISSUE #242: Puller: docs not inserted when some docs fail validation due to nested transaction behavior
From [google groups](https://groups.google.com/d/msgid/mobile-couchbase/ea5110f9-19bc-4eb7-b0f9-4570dabc17b3%40googlegroups.com?utm_medium=email&utm_source=footer) discussion:

> This means that if I do a Pull replication and have to process about 100 different documents, if the inner transaction that deals with each document fails validation -- even if it is one document -- then the entire 100 documents are not inserted into the local database. That seems like a bug to me since the outer transaction could have documents that pass the validation and hence should be available to the mobile.

hideki:
Reason behind this implementation
1. `SQLiteDatabase.beginTransaction()` supported nested transaction. But, in case, nested transaction rollbacked, it also rollbacks outer transaction too. This is not ideal behavior for CBL
2. `SAVEPOINT...RELEASE` supports nested transaction. But Android API 14 and 15, it throws Exception. I assume it is Android bug. As CBL need to support from API 10. So it does not work for CBL.
3. Using Transaction for outer/1st level of transaction and inner/2nd level of transaction works with CBL requirement.
4. CBL Android and Java uses Thread, So it is better to use `SQLiteDatabase.beginTransaction()` for outer/1st level transaction. if we use `BEGIN...COMMIT` and `SAVEPOINT...RELEASE`, we need to implement wrapper of `BEGIN...COMMIT` and `SAVEPOINT...RELEASE` to be Thread-safe.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -1228,7 +1228,7 @@ public final class Database {
             }
             // Inner (level 1 or higher) transaction. Use SQLite's SAVEPOINT
             else{
-                database.execSQL("SAVEPOINT tdb" + Integer.toString(transactionLevel));
+                database.execSQL("SAVEPOINT cbl_" + Integer.toString(transactionLevel));
             }
 
             Log.i(Log.TAG, "%s Begin transaction (level %d)", Thread.currentThread().getName(), transactionLevel);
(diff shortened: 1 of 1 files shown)


hideki:
@pasin Could you please review this again?

PR #460, opened 2015-02-24
Fixed #458 - Upgrade Jackson to v2.5.0

### LINKED ISSUE #458: Upgrade Jackson to latest (2.5.0)
Current version of CBL Android uses Jackson 1.9.2. 
Latest version of Jackson is 2.5.0. (http://jackson.codehaus.org/)

hideki:


tleyden:
@hideki can you try to build on the jenkins machine and make sure it compiles there and there are no new unit test failures?

pasin:
@tleyden Can you help reviewing this PR? I don't know the history of using Jackson with SQLite blob (byte arrays with zero terminated).

hideki:
http://172.23.113.52:8080/job/couchbase-lite-android-build/46/

hideki:
It seems that Jenkins build and unit tests passed.

src/main/java/com/couchbase/lite/support/JsonDocument.java
```diff
@@ -36,7 +36,13 @@ public Object jsonObject() {
                 tmp = new LazyJsonArray<Object>(json);
             } else {
                 try {
-                    tmp = Manager.getObjectMapper().readValue(json, Object.class);
+                    // this is for Jackson 2.5.0
```
tleyden: I'm confused about what this is doing here.. why is this needed?

hideki: json is byte[] which is from `Cursor.getBlob()`. And json byte array is ended with '\0'. '\0' causes parsing problem with Jackson 2.5.0. We did not observe this problem with Jackson 1.9.2.

tleyden: ok that makes sense.  can you add that to the comments?

so does that mean we are giving invalid json to Jackson?

hideki: - Added comment, and pushed it.
- As byte[] has length property, it does not require NULL stop like C char array. So It seems Jackson 2.5.0 treats '\0' as part of data, but it is not able to handle it. I believe `Cursor.getBlob()` does not need to added '\0' at end of byte[], also Jackson 2.5.0 should handle it.

PR #519, opened 2015-03-25
Using request headers for multipart request

reisub:
The multipart requests don't include headers set for the replication, this commit fixes the problem.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/RemoteMultipartRequest.java
+++ b/src/main/java/com/couchbase/lite/support/RemoteMultipartRequest.java
@@ -42,6 +42,7 @@ public class RemoteMultipartRequest extends RemoteRequest {
             throw new IllegalArgumentException("Invalid request method: " + method);
         }
 
+        addRequestHeaders(request);
         request.addHeader("Accept", "*/*");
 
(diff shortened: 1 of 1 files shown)


hideki:
Hi @reisub 

We have a couple of requests before merging your PR.
1. Could you please sign [Sign the contributor license agreement](https://github.com/couchbase/couchbase-lite-ios/wiki/Sign-the-contributor-license-agreement)
1. Could you please explain the problem?

Thanks!

reisub:
Hi @hideki,
I already signed the CLA for an earlier PR: https://github.com/couchbase/couchbase-lite-java-core/pull/298

CBL replication has a `setHeaders()` method which takes a map of headers and makes the replication add additional headers to each http request it makes. This works for every type of request we use, but not for a multipart request. We traced the problem back to this class (`RemoteMultipartRequest`) where these headers are not used at all, although they are passed in via the constructor. This change fixes that.

hideki:
Hi @reisub, 

Thank you very much for your help! Some reason we forgot to call `addRequestHeaders(request);` from RemoteMultipartRequest.

Cheers!

PR #564, opened 2015-04-13
UnsavedRevision: Preserve insertion order.

sergio91pt:
Preserve insertion order in the JSON by using LinkedHashMap's.
This is also used by Jackson [1] and the .NET port, it seems [2].

Altough not really a necessity, preserving insertion order is a nice to have,
helpeful when someone wants to compare multiple JSON documents.

[1]: http://wiki.fasterxml.com/JacksonDataBinding#line-54
[2]: https://github.com/couchbase/couchbase-lite-net/blob/master/src/Couchbase.Lite.Shared/Sharpen/LinkedHashMap.cs

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -42,6 +42,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.Object;
+import java.lang.String;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.Charset;
@@ -53,6 +55,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -1576,7 +1579,7 @@ public final class Database {
             revsInfo = new ArrayList<Object>();
             List<RevisionInternal> revHistoryFull = getRevisionHistory(rev);
             for (RevisionInternal historicalRev : revHistoryFull) {
-                Map<String,Object> revHistoryItem = new HashMap<String,Object>();
+                Map<String,Object> revHistoryItem = new LinkedHashMap<String,Object>();
                 String status = "available";
                 if(historicalRev.isDeleted()) {
                     status = "deleted";
@@ -1605,7 +1608,7 @@ public final class Database {
             }
         }
 
-        Map<String,Object> result = new HashMap<String,Object>();
+        Map<String,Object> result = new LinkedHashMap<String, Object>();
         result.put("_id", docId);
         result.put("_rev", revId);
         if(rev.isDeleted()) {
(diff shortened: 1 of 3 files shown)


hideki:
Hi @sergio91pt 

We have a couple of requests before merging your PR.
1. Could you please sign Sign the contributor license agreement  => https://github.com/couchbase/couchbase-lite-ios/wiki/Sign-the-Contributor-License-Agreement
2. Could you please write simple UnitTest? => https://github.com/couchbase/couchbase-lite-android/blob/master/src/androidTest/java/com/couchbase/lite/DocumentTest.java

Thanks!

hideki:
Hi @sergio91pt,

Couchbase Lite support multiple platforms iOS, Android (Java), and .NET. We tried them to identical as much as possible. Currently we are discussing if applying LinkedHashMap is good way.

- I talked with .NET developer for Couchbase Lite. .NET version does not use LikedHashMap for UnsavedRevision.
- Is there specific use-case for preserving insertion order?

Thanks!

sergio91pt:
I saw the LinkedHashMap in the .NET tree and did not investigate further, thanks for clarifying that.

The main advantage is ease on the eyes, as it is nicer for documents to mostly have the same order on their properties when someone is looking through some of them on Fuxton, for example.

Right now the order is totally dependent on the HashMap implementation (object hashcode + secondary hashcode) and may change the order of all properties when a document is updated on a different device or Android version (which would diff's of revisions harder to read).

It's easier to catch problems on the documents, if the order is mostly the same and the performance of a LinkedHashMap hit is minimal. I suppose Jackson also uses a LinkedHashMap due to a similar rationale.

Either way this patch still needs some work as RevisionInternal changes the order on the second revision and then there's all the Attachment stuff on Database.java.

sergio91pt:
The PR is ready for review and with unit tests. However I can't commit to port the changes to the iOS and .NET implementations.

hideki:
Hi @sergio91pt,

Using LinkedHashMap is good idea to preserve the insertion order. However, with following reasons, applying LinkedHashMap to UnsavedRevision is not good enough to preserver the insertion order.

- There are many codes that use HashMap. Therefore, using LinkedHashMap for UnsavedRevision can not preserve insertion order at all cases.
- Sync Gateway and CouchDB do (might) not preserve insertion order of JSON/Document. Using LinkedHashMap for CBL Android/Java can not solve all scenarios.
- Also CBL iOS and .NET do not preserve insertion order. By sync document with iOS/.NET, insertion order of document is not preserved.
- Ideally, we would like to sort entries in a document by canonical order using key.

We would like to make this PR as pending state.

Again, Thank you very much for your contributions!

hideki:
Hi @sergio91pt,
In short term, we will not apply this change because it addresses part of problems. For now, I will close this PR.
Thank you for your contribution!

sergio91pt:
@hideki  

> Sync Gateway and CouchDB do (might) not preserve insertion order of JSON/Document. Using LinkedHashMap for CBL Android/Java can not solve all scenarios.

RCouch preserves the order it is given to but adds the _id and _rev to the top. CouchDB probably does the same.

> Ideally, we would like to sort entries in a document by canonical order using key.

Regarding this part of your previous comment, adding a small abstraction around ObjectMapper would be probably the way to achieve this, as it would ease testing other implementations #234 (because of speed, size and/or method count), and would allow to due such sorting in a centralized place. 

Jackson even has a config setting to make the serializer order props alphabetically (`SerializationConfig#SORT_PROPERTIES_ALPHABETICALLY`), using TreeMaps.
This way the developer could choose to use a different parser/serializer and implement a custom ordering (if it matters enough given the overhead).

> In short term, we will not apply this change because it addresses part of problems. For now, I will close this PR.

Sure, I understand. Does't make sense to keep it open, if canonical order is the way to go...

hideki:
Hi @sergio91pt, Thank you for your comments. I always learn from you a lot.

PR #566, opened 2015-04-14
Fixed #419 (multiple document updates in single transaction)

### LINKED ISSUE #419: Allow a doc to be updated more than once in a single transaction
Should be ported from upstream ios: https://github.com/couchbase/couchbase-lite-ios/pull/589

hideki:
- Cache.java, DocumentChange.java and RevisionInternal.java - implemented missing methods
- Document.java - updated revisionAdded() method, and implemented missing method
- Database.java - Implemented storageExitedTransaction and databaseStorageChanged to notify document change during transaction, and also implemented missing method

### DIFF
--- a/src/main/java/com/couchbase/lite/Cache.java
+++ b/src/main/java/com/couchbase/lite/Cache.java
@@ -76,6 +76,7 @@ public class Cache<K,V> {
         weakReferenceCache.clear();
     }
 
-
-
+    public V resourceWithCacheKeyDontRecache(K key) {
+        return weakReferenceCache.get(key);
+    }
 }
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -3639,7 +3639,7 @@ public final class Database {
         }
 
         // Squish the change objects if too many of them are piling up:
-        if (changesToNotify.size() > MANY_CHANGES_TO_NOTIFY) {
+        if (changesToNotify.size() >= MANY_CHANGES_TO_NOTIFY) {
             if(changesToNotify.size() == MANY_CHANGES_TO_NOTIFY){
                 for(DocumentChange c : changesToNotify){
                     c.reduceMemoryUsage();
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -72,6 +72,9 @@ public final class Database {
     // Default value for maxRevTreeDepth, the max rev depth to preserve in a prune operation
     private static final int DEFAULT_MAX_REVS = Integer.MAX_VALUE;
 
+    // When this many changes pile up in _changesToNotify, start removing their bodies to save RAM
+    private static final int MANY_CHANGES_TO_NOTIFY = 5000;
+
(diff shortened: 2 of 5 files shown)


sergio91pt:
I'm not sure as I'm not familiar enough with the codebase to review this PR, but seems it exposes revisions not yet committed to other threads (MySQL transactions are thread local)

I made [this test](https://github.com/sergio91pt/couchbase-lite-android/commit/ae76cb61b3e2a48ab63383c955b5b0bd164f236a) and it fails on this branch but not on master.

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -3610,50 +3608,63 @@ private void postChangeNotifications() {
...
+            // TODO: change this to match iOS and call cachedDocumentWithID
+            //cachedDocumentWithID(change.getDocumentId()).revisionAdded(change, false);
+            getDocument(change.getDocumentId()).revisionAdded(change, false);
+        }
 
-    private void notifyChanges(List<DocumentChange> documentChanges) {
-        if (changesToNotify == null) {
-            changesToNotify = new ArrayList<DocumentChange>();
+        // Squish the change objects if too many of them are piling up:
+        if (changesToNotify.size() > MANY_CHANGES_TO_NOTIFY) {
```
pasin: Should it be >= ?

src/main/java/com/couchbase/lite/Cache.java
```diff
@@ -76,6 +76,7 @@ public void clear() {
         weakReferenceCache.clear();
     }
 
-
-
+    public V resourceWithCacheKeyDontRecache(K key) {
```
pasin: Are we using this anywhere or plan to use this?

pasin:
I believed iOS has only one (only one level)  document cache. 

1. It's a weak smart cache that if a document inside a cache doesn't get retained by any ones (e.g. reference count = 0), the document will get automatically removed from the cache. Otherwise, it will stay inside the cache. So in normal case, if a document is still get used or referenced by something, getting the document from the cache should not be null.

2. Inside the cache itself, it also makes strong references to documents up to 50 (recently-used) docs so those 50 recently used docs get kept inside the cache even though they are not get retained by any other objects outside the cache.

3. When the database.purgeDocument: method is called, the purged document will get explicitly removed from the cache.

hideki:
Hi @sergio91pt,
Thank you very much for reviewing the PR.
CBL Android uses SQLiteDatabase to access SQLite. SQLiteDatabase.beginTransaction() ( http://developer.android.com/reference/android/database/sqlite/SQLiteDatabase.html#beginTransaction() ) is a thread safe.

src/main/java/com/couchbase/lite/Database.java
hideki: @pasin Thanks a lot.

src/main/java/com/couchbase/lite/Cache.java
hideki: This is equivalent implementation of `- (id<CBLCacheable>) resourceWithCacheKeyDontRecache: (NSString*)cacheKey;` in CBLCache.h/.m. I'd like to keep it.

PR #567, opened 2015-04-14
BlobStoreWriter: Throw IOExceptions.

sergio91pt:
Instead of throwing an unchecked exception and ignoring the close on the
output file, throw checked IOExceptions.

We should consider an fsync before closing the outputstream to get a better
consistency guarantee.

With this commit, when saving an attachment fails on UnsavedRevison#save
a CBLException is thrown with STATUS_ATTACHMENT_ERROR.

MultipartDocumentReader continues to throw unchecked exceptions.


sergio91pt:
Unit tests in couchbase/couchbase-lite-android#576

hideki:
@sergio91pt Thank you very much for your contribution!

PR #568, opened 2015-04-15
Encode userinfo in BasicAuthenticator.

sergio91pt:
Basic authentication was broken if the username contained a ':'.

### DIFF
--- a/src/main/java/com/couchbase/lite/auth/BasicAuthenticator.java
+++ b/src/main/java/com/couchbase/lite/auth/BasicAuthenticator.java
@@ -1,5 +1,7 @@
 package com.couchbase.lite.auth;
 
+import com.couchbase.lite.util.URIUtils;
+
 import java.net.URL;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


hideki:
Hi @sergio91pt,

CBL Android/Java need to be compatible with CBL iOS and .NET.
UTF-8 encode then Base64 encode is right approach for Couchbase Lite products.
https://github.com/couchbase/couchbase-lite-ios/blob/b3b64f38b6278f39b5e90362906f443100e9f536/Source/CBLAuthorizer.m#L43-L54

Thanks,
Hideki

sergio91pt:
Hi @hideki, sorry for wasting your time but I'm going to drop this PR.
According to rfc 1945 usarnames using basic authentication cannot contain a ':', so there's no point in escaping the userinfo part, if it is not going to be included in an actual URL.

hideki:
Hi @sergio91pt,
I could learn a lot from your PRs and comments. 
Thanks a lot!

PR #569, opened 2015-04-16
Fix authinfo with non-ASCII passwords.

sergio91pt:
Fixes couchbase/couchbase-lite-android#382

I tested this with the performance.Test7_PullReplication testcases and wireshark.
To test this either [percent encode](http://coderstoolbox.net/string/) the username and password (where needed) or use couchbase/couchbase-lite-android#582.

I used wireshark to check the response for the GET /db/_local/uuid, because the test does not fail if unauthorized.

Tested on rcouch using with 七乃又直ந்த as the username, and ؤلمن:يстекло as the password. Of course the account must be created beforehand.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/CouchbaseLiteHttpClientFactory.java
+++ b/src/main/java/com/couchbase/lite/support/CouchbaseLiteHttpClientFactory.java
@@ -3,6 +3,7 @@ package com.couchbase.lite.support;
 import com.couchbase.lite.Database;
 import com.couchbase.lite.internal.InterfaceAudience;
 
+import org.apache.http.auth.params.AuthPNames;
 import org.apache.http.client.CookieStore;
 import org.apache.http.client.HttpClient;
 import org.apache.http.conn.ClientConnectionManager;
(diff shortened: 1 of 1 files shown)


hideki:
Hi @sergio91pt,

Did you have chance to test against sync_gateway?

sergio91pt:
@hideki I didn’t because the only dev env I've setup is rcouch 1.1.0, however I checked with wireshark to make sure the password was correctly encoded on the wire.

If the charset is left unset it [defaults to ASCII](http://androidxref.com/5.1.0_r1/xref/external/apache-http/src/org/apache/http/auth/params/AuthParams.java), while the [ios impelementation](https://github.com/couchbase/couchbase-lite-ios/blob/b3b64f38b6278f39b5e90362906f443100e9f536/Source/CBLAuthorizer.m#L43-L54) you pointed in 568 encodes using UTF-8.

hideki:
Hi @sergio91pt,
I will test this PR early next week, and I will merge if no issue with PR.
Thanks!

hideki:
@sergio91pt ,
Thank you very  much for your contribution!

PR #574, opened 2015-04-17
Fixed #456 - Sync reports being in Idle state when it is still in the middle of syncing

### LINKED ISSUE #456: Sync reports being in Idle state when it is still in the middle of syncing
Greetings,

Sync will report being in an idle state with 0/0 documents being synced in the middle of syncing. For example:

02-24 10:48:46.312    2856-2890/com.app.app.alpha D/App﹕ Progress: 149 / 586
02-24 10:48:46.312    2856-2890/com.app.app.alpha D/App﹕ ----------->Sync Session: active: true status: Active 149/586 Error (null)
02-24 10:48:46.382    2856-2890/com.app.app.alpha D/App﹕ Progress: 0 / 0
02-24 10:48:46.382    2856-2890/com.app.app.alpha D/App﹕ ----------->Sync Session: active: false status: Idle 0/0 Error (null)

This makes it impossible to know when an initial sync is complete and doesn't mirror the iOS behavior.

Cheers,
Justin

hideki:
- `PullerInternal.change()` creates thread which trigger IDEL whenever it is called. Prevent to multiple times to start thread by flag
- To wait all queue become empty, use while loop instead of just sequential queue check.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -838,28 +838,38 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
                 @Override
                 public void run() {
                     try {
-                        if (batcher != null) {
-                            // if batcher delays task execution, need to wait same amount of time. (0.5 sec or 0 sec)
-                            try { Thread.sleep(batcher.delayToUse()); }catch(Exception e){}
-                            Log.e(Log.TAG_SYNC, "batcher.waitForPendingFutures()");
-                            batcher.waitForPendingFutures();
-                        }
+                        // NOTE: Wait till all queue becomes empty
+                        while ((batcher != null && batcher.count() > 0) ||
+                                (pendingFutures != null && pendingFutures.size() > 0) ||
+                                (downloadsToInsert != null && downloadsToInsert.count() > 0)) {
 
-                        Log.e(Log.TAG_SYNC, "waitForPendingFutures()");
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/PullerInternal.java
```diff
@@ -880,45 +877,87 @@ protected void stopGraceful() {
...
+            try {
+                waitForAllTasksCompleted();
+            } catch (Exception e) {
+                Log.e(Log.TAG_SYNC, "Exception waiting for pending futures: %s", e);
+            } finally {
+                Log.d(Log.TAG_SYNC, "[waitForPendingFutures()] END - thread id: " + Thread.currentThread().getId());
+                fireTrigger(ReplicationTrigger.WAITING_FOR_CHANGES);
+                waitingForPendingFutures = false;
+            }
+        }
```
pasin: I'm not sure that I fully understand the purpose of this synchronized block. Do you expect multiple threads (if any) to be blocked until waitForAllTasksCompleted() is done? or Do you want to lock multiple access to waitingForPendingFutures variable?

hideki: `waitForPendingFutures ()` should not be called by multiple threads. However, because of multiple threading, there is risk that `waitForPendingFutures ()` is called by multiple threads during waiting the task completed.

With synchronized block, the code makes sure `waitForPendingFutures ()` are executed by only one thread at same time.

PR #578, opened 2015-04-22
Allow allDocumentsQuery().toLiveQuery().

sergio91pt:
Fixes couchbase/couchbase-lite-android#237
Also fixes an infinite loop in waitForRows(), if another thread calls stop().


hideki:
@sergio91pt Thank you very much!

PR #582, opened 2015-04-23
Fixed #456 - Sync reports being in Idle state when it is still in the middle of syncing

### LINKED ISSUE #456: Sync reports being in Idle state when it is still in the middle of syncing

hideki:
Instead of merging into master, merge into release/1.1.0 branch.

Previous PR: https://github.com/couchbase/couchbase-lite-java-core/pull/574

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -903,23 +903,27 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
     protected Object lockWaitForPendingFutures = new Object();
 
     public void waitForPendingFutures() {
-        if (waitingForPendingFutures) {
-            return;
-        }
         synchronized (lockWaitForPendingFutures) {
-            waitingForPendingFutures = true;
-
-            Log.d(Log.TAG_SYNC, "[PullerInternal.waitForPendingFutures()] STARTED - thread id: " + Thread.currentThread().getId());
-
-            try {
-                waitForAllTasksCompleted();
-            } catch (Exception e) {
-                Log.e(Log.TAG_SYNC, "Exception waiting for pending futures: %s", e);
-            } finally {
-                Log.d(Log.TAG_SYNC, "[waitForPendingFutures()] END - thread id: " + Thread.currentThread().getId());
-                fireTrigger(ReplicationTrigger.WAITING_FOR_CHANGES);
-                waitingForPendingFutures = false;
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/PullerInternal.java
```diff
@@ -880,45 +877,87 @@ protected void stopGraceful() {
...
+            try {
+                waitForAllTasksCompleted();
+            } catch (Exception e) {
+                Log.e(Log.TAG_SYNC, "Exception waiting for pending futures: %s", e);
+            } finally {
+                Log.d(Log.TAG_SYNC, "[waitForPendingFutures()] END - thread id: " + Thread.currentThread().getId());
+                fireTrigger(ReplicationTrigger.WAITING_FOR_CHANGES);
+                waitingForPendingFutures = false;
+            }
+        }
```
pasin: From your comment on the previously closed PR, you would like to block so that there will be only one thread executing this code at a time. If I understand correctly, you don't need waitingForPendingFutures variable and the if statement above the synchronized block. Even put synchronized at the method level should achieve your goal. 

Currently there are 2 different scenarios that could happen depending on racing condition:
1. The "if (waitingForPendingFutures)" outside the synchronized block means there is no needs to execute this method  when one of thread is executing.
2. Some threads may get blocked at the synchronized block.

hideki: `waitingForPendingFutures` variable is required for prevent unnecessary thread calls `fireTrigger(ReplicationTrigger.WAITING_FOR_CHANGES);` at line 920. This makes a replicator IDLE. Without checking `waitingForPendingFutures`, calling `waitForPendingFutures()` method from all threads calls `fireTrigger(ReplicationTrigger.WAITING_FOR_CHANGES);`.

PR #597, opened 2015-04-25
Fixed #575 repl restart hang

### LINKED ISSUE #575: restart of a replication shouldn't hang if the replication is already stopped.
Calling restart on a replication will hang if the replication is already stopped.  Even if you check before calling restart, there is a race if the replicator stops between the check and the establishment of the listener inside the restart method.

Below is the code in Replication.java.  The restart method needs to check if the replicator is already  stopped before waiting on the countdown latch.

```
    public void restart() {

        final CountDownLatch stopped = new CountDownLatch(1);
        addChangeListener(new ChangeListener() {
            @Override
            public void changed(ChangeEvent event) {
                if (event.getTransition() != null && event.getTransition().getDestination() == ReplicationState.STOPPED) {
                    stopped.countDown();
                }
            }
        });

        stop();

        try {
            stopped.await();
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }

        start();
}
```

hideki:


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -194,26 +195,34 @@ public void start() {
...
-        addChangeListener(new ChangeListener() {
-            @Override
-            public void changed(ChangeEvent event) {
-                if (event.getTransition() != null && event.getTransition().getDestination() == ReplicationState.STOPPED) {
-                    stopped.countDown();
+        // stop replicator if necessary
+        if(this.isRunning()) {
+            final CountDownLatch stopped = new CountDownLatch(1);
+            addChangeListener(new ChangeListener() {
+                @Override
```
pasin: Should remove the listener after the restart is done too.

hideki:
Hi @pasin,
Applied your feedback. Please review PR again.
Thanks!

PR #599, opened 2015-04-25
Fix get attachment stream fails on UnsavedRevision

### LINKED ISSUE #592: Retrieving attachment data fails on UnsavedRevision
Attachment#getContent() always fails to retrieve the data set on a previous revision (eg. the parent revision), if the Attachment object is retrieved through an UnsavedRevision.

The problem is the sequence number that is always 0 on UnsavedRevision and the parent sequence is not used. Should an UnsavedRevision even have a sequence number?

```
Caused by: com.couchbase.lite.CouchbaseLiteException
 	at com.couchbase.lite.Database.getAttachmentForSequence(Database.java:2820)
 	at com.couchbase.lite.Attachment.getContent(Attachment.java:130)
```

sergio91pt:
Closes #592.

This introduces a new package protected method getParentSequence().
The alternative was to call getParent().getSequence() which leads to an extra
database query everytime an attachment is retrieved from a non-current revision.

### DIFF
--- a/src/main/java/com/couchbase/lite/Attachment.java
+++ b/src/main/java/com/couchbase/lite/Attachment.java
@@ -28,6 +28,7 @@ import java.net.URL;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.zip.GZIPInputStream;
 
 /**
  * A Couchbase Lite Document Attachment.
@@ -132,6 +133,16 @@ public final class Attachment {
             }
             Attachment attachment = db.getAttachmentForSequence(sequence, this.name);
             body = attachment.getContent();
+            if (attachment.getGZipped()) {
+                // Client does not expect a gzipped stream.
+                // Only Router handles gzipped streams and uses getAttachmentForSequence directly.
+                try {
+                    body = new GZIPInputStream(body);
+                } catch (IOException e) {
+                    throw new CouchbaseLiteException(e.getMessage(), Status.STATUS_ATTACHMENT_ERROR);
+                }
+            }
+            gzipped = false;
             return body;
         }
     }
--- a/src/main/java/com/couchbase/lite/Attachment.java
+++ b/src/main/java/com/couchbase/lite/Attachment.java
@@ -124,13 +124,24 @@ public final class Attachment {
     public InputStream getContent() throws CouchbaseLiteException {
         if (body != null) {
             return body;
-        }
-        else {
+        } else {
             Database db = revision.getDatabase();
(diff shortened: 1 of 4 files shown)


sergio91pt:
Unite tests: https://github.com/couchbase/couchbase-lite-android/issues/592

sergio91pt:
Hi @hideki,
I added a commit to automatically decode gzipped attachments for Revision users (see commit message). I did not create another PR as it would probably conflict with this one.

Back on topic, CBL .NET [upgraded](https://github.com/couchbase/couchbase-lite-net/commit/6144a4a11fd9b867d70cc30d8e763a75d4c0f047) their `_digest` to use sha1 and so they avoid using getAttachmentForSequence to retrieve the attachment key (in all cases - rev and unsaved rev).

I'm not sure how they handled backward compatibility: I see code for putRevision but no code to handle the retrieval of documents still with md5 digests. (Regarding MultiparDocumentReader, it still uses md5 despite that variable name change).

We could also use sha1 digests and fallback to getAttachmentForSequence if the digest was md5.

hideki:
Hi @sergio91pt,
Give me few days to review the PR. I have some other tasks to finish.
Thanks again for your contributions!
Hideki

src/main/java/com/couchbase/lite/Attachment.java
```diff
@@ -124,15 +125,36 @@ public String getContentType() {
...
+        } else {
             Database db = revision.getDatabase();
-            Attachment attachment = db.getAttachmentForSequence(revision.getSequence(), this.name);
+            long sequence = getAttachmentSequence();
+            if (sequence == 0) {
+                throw new CouchbaseLiteException(Status.INTERNAL_SERVER_ERROR);
+            }
+            Attachment attachment = db.getAttachmentForSequence(sequence, this.name);
             body = attachment.getContent();
+            if (attachment.getGZipped()) {
```
hideki: As you commented in the code, Router could expects gzipped file. Need to update this logic.
https://github.com/couchbase/couchbase-lite-java-core/blob/12c29af1b4f05aa260bfa36c6a02913d92acb47e/src/main/java/com/couchbase/lite/router/Router.java#L1646-L1681

sergio91pt: Router uses Databse.getAttachmentForSequence to retrieve the attachment which uses the constructor with the InputStream. This else clause is only reached when instantiated through a Revision.

hideki: As you wrote, `body` instance variable should not be null in case the instance is created from `Database.getAttachmentForSequence`. Thanks.

hideki:
Thanks again for your contributions!

PR #606, opened 2015-04-29
Fixed #596 - Handle SQL 'attach' table incompatibility in CBL 1.1

### LINKED ISSUE #596: Handle SQL 'attach' table incompatibility in CBL 1.1
Per @snej's guidance:
```
detect whether a database was created on Obj-C-1.1 and if so recreate the ‘attach’ table.

When opening a database, check whether the ‘attach’ table exists. If not:
Create the ‘attach’ table.
Iterate over every revision containing attachments (rows in ‘revs’ whose ‘noattach' column is not 1.)
Parse the revision’s JSON and get the ‘_attachments’ property.
For every item in that map, create a row in the ‘attach’ table. (It may be possible to reuse or at least copy existing Database code, since this is the same thing that happens when adding a revision to the database.)
```

hideki:
- Check dbVersion. If >= 200, current version can not support
- If dbVersion >= 100, Check if attachments table exists. If not, create attachments table, and iterates revs table to populate attachments table
- Set current database version 20.  We should use the number less than 101 and larger than previous released version

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -53,6 +53,7 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -997,7 +998,7 @@ public final class Database {
         int dbVersion = database.getVersion();
 
         // Incompatible version changes increment the hundreds' place:
-        if(dbVersion >= 100) {
+        if(dbVersion >= 200) {
             Log.e(Database.TAG, "Database: Database version (%d) is newer than I know how to work with", dbVersion);
             database.close();
             return false;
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1210,6 +1211,111 @@ public synchronized boolean open() {
...
+                    }
+                    database.close();
+                    return false;
+                } finally {
+                    if (cursor2 != null) {
+                        cursor2.close();
+                    }
+                }
+            }
+            dbVersion = 20;
```
pasin: Do you need to change the user_version to 20 also?

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1210,6 +1211,111 @@ public synchronized boolean open() {
...
+                        cursor2.close();
+                    }
+                    database.close();
+                    return false;
+                } finally {
+                    if (cursor2 != null) {
+                        cursor2.close();
+                    }
+                }
+            }
```
pasin: I could be faster if you are using transaction.

src/main/java/com/couchbase/lite/Database.java
pasin: I see. You have "PRAGMA user_version = 20" in the upgrade sql.

src/main/java/com/couchbase/lite/Database.java
hideki: We will do it after 1.1 release. 
Here is ticket for it https://github.com/couchbase/couchbase-lite-java-core/issues/607

src/main/java/com/couchbase/lite/Database.java
pasin: I think if there is an error while migrating, we should rollback everything (including the created attachments table and schema version).

PR #621, opened 2015-05-13
Fixed CBL Android 197 gzipped attachment

hideki:
- Fixed `Attachment.getContent()`. It causes recursive calls that does unzip twice.
- Fixed attachments table access. Previous codes does not handle properly `encoding` and `encoding_length` columns.
- Fixed storing Basic64 encoded gzipped attachment. It was unzipped before storing attachment.
- Fixed push replication to set `Content-Encoding` for gzipped attachment

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -3159,16 +3159,16 @@ public final class Database {
                 }
 
                 String encodingStr = null;
-                if(encoding!=AttachmentInternal.AttachmentEncoding.AttachmentEncodingNone){
+                if (encoding == AttachmentInternal.AttachmentEncoding.AttachmentEncodingGZIP) {
                     // NOTE: iOS decode if attachment is included int the dict.
                     encodingStr = "gzip";
                 }
 
                 Map<String, Object> attachment = new HashMap<String, Object>();
-                if(!(dataBase64 != null || dataSuppressed)) {
+                if (!(dataBase64 != null || dataSuppressed)) {
                     attachment.put("stub", true);
                 }
-                if(dataBase64 != null) {
+                if (dataBase64 != null) {
                     attachment.put("data", dataBase64);
                 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -3140,39 +3149,40 @@ public boolean sequenceHasAttachments(long sequence) {
...
                         }
                         else {
                             Log.w(Database.TAG, "Error loading attachment.  Sequence: %s", sequence);
                         }
-
                     }
+                }
 
+                String encodingStr = null;
+                if(encoding!=AttachmentInternal.AttachmentEncoding.AttachmentEncodingNone){
```
pasin: I knew now there is only None and GZip. It is a little safer to check against Gzip.

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -2882,37 +2896,28 @@ public void insertAttachmentForSequenceWithNameAndType(InputStream contentStream
...
             args.put("revpos", revpos);
+            if (key != null) {
+                args.put("key", key.getBytes());
+            }
+            if (length >= 0) {
+                args.put("length", length);
+            }
+            if(encoding == AttachmentInternal.AttachmentEncoding.AttachmentEncodingGZIP) {
+                args.put("encoding", encoding.ordinal());
+                if (encodedLength >= 0) {
```
pasin: Should we have an assert statement here to ensure that there is no case that the encodedLength < 0 when gzip encoding is specified?

hideki: encodedLength is information only. In case encodedLength is less than 0, the code does not store the value. It will set null value in encoded-length column in SQLite attachments table.

src/main/java/com/couchbase/lite/Database.java
hideki: Yes. I will fix it.

PR #668, opened 2015-06-22
Fix for #667: changed stream implementations in URLConnection

### LINKED ISSUE #667: Using PipedInputStream and PipedOutputStream in URLConnection can cause deadlock (see issue 57 in couchbase-lite-java-listener)
The URLConnection class uses PipedInputStream and PipedOutputStream. This can cause a deadlock when writing and reading on the same thread (see issue 57 in couchbase-lite-java-listener).

fpum:
* fixes #667 
* changed stream implementations in URLConnection to custom ones with a resizable buffer

### DIFF
--- a/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
+++ b/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
@@ -2,7 +2,6 @@ package com.couchbase.lite.router;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.LinkedList;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class BufferOutputStream extends OutputStream {
@@ -14,6 +13,7 @@ public class BufferOutputStream extends OutputStream {
         if(!isClosed()) {
             synchronized (buffer) {
                 buffer.push((byte) i);
+                buffer.notify();
             }
         } else {
             throw new IOException("Can't write to closed stream.");
@@ -26,6 +26,9 @@ public class BufferOutputStream extends OutputStream {
 
     public void close() {
         closed.set(true);
+        synchronized(buffer) {
+            buffer.notify();
+        }
     }
 
     public boolean isClosed() {
--- a/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
+++ b/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
@@ -5,9 +5,6 @@ import java.io.OutputStream;
 import java.util.LinkedList;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-/**
- * Created by florian on 19.06.2015.
- */
 public class BufferOutputStream extends OutputStream {
 
     LinkedList<Byte> buffer = new LinkedList<Byte>();
--- a/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
+++ b/src/main/java/com/couchbase/lite/router/BufferOutputStream.java
@@ -1,0 +1,34 @@
+package com.couchbase.lite.router;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.LinkedList;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Created by florian on 19.06.2015.
+ */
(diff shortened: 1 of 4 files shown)


hideki:
Hi @fpum,
Thank you very much for your contribution.
Have you ever signed our contributor license agreement? If not, could you please sign it?
https://github.com/couchbase/couchbase-lite-ios/wiki/Sign-the-contributor-license-agreement
Thanks a lot!

src/main/java/com/couchbase/lite/router/URLConnection.java
```diff
@@ -47,12 +45,8 @@
...
     public URLConnection(URL url) {
         super(url);
-        responseInputStream = new PipedInputStream();
-        try {
-            responseOutputStream = new PipedOutputStream((PipedInputStream)responseInputStream);
-        } catch (IOException e) {
-            Log.e(Log.TAG_ROUTER, "Exception creating piped output stream", e);
-        }
+        responseOutputStream = new BufferOutputStream();
+        responseInputStream = new BufferInputStream((BufferOutputStream)responseOutputStream);
```
hideki: I understand `PipedInputStream` and `PipedOutputStream` causes deadlock if they are used from single thread.

Can we just use [BufferedInputStream](http://docs.oracle.com/javase/6/docs/api/java/io/BufferedInputStream.html) and [BufferedOutputStream](http://docs.oracle.com/javase/6/docs/api/java/io/BufferedOutputStream.html) instead of writing `BufferInputStream` and `BufferOutputStream `?

fpum:
Hi @hideki,
I signed the contribution agreement.

src/main/java/com/couchbase/lite/router/URLConnection.java
fpum: The BufferedOuputStream from the java.io package just adds a buffer of fixed size to any other output stream. This means it will write data into its internal buffer and write the contents of the buffer to the underlying stream when its buffer is full, the flush method is called or the stream gets closed. If writing to the underlying stream blocks, this will still result in a deadlock.

hideki:
Hi @fpum,
Thank you for signing the agreement.

src/main/java/com/couchbase/lite/router/BufferOutputStream.java
```diff
@@ -0,0 +1,31 @@
+package com.couchbase.lite.router;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.LinkedList;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class BufferOutputStream extends OutputStream {
+
+    LinkedList<Byte> buffer = new LinkedList<Byte>();
```
hideki: Using `LinkedList<Byte>` for storing data is not efficient from both memory usage and performance.

src/main/java/com/couchbase/lite/router/BufferOutputStream.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public void close() {
+        closed.set(true);
+    }
+
+    public boolean isClosed() {
+        return closed.get();
+    }
+
```
hideki: Needs to override [write(byte[], int, int)](http://docs.oracle.com/javase/7/docs/api/java/io/OutputStream.html#write(byte[],%20int,%20int)) also.

src/main/java/com/couchbase/lite/router/BufferOutputStream.java
```diff
@@ -0,0 +1,31 @@
...
+}
```
hideki: I believe Byte by Byte operation with using LinkedList<Byte> is not efficient.

src/main/java/com/couchbase/lite/router/BufferOutputStream.java
fpum: I don't think that it much of a perfromance issue since it is only appending at the end and removing from the first position. But you are right, an array would be more memory efficient.

hideki:
Hi @fpum,
Sorry for taking long time to merge your PR. And Thank you very much for your PR and test code. Your test codes helped us a lot.
Thanks!

PR #701, opened 2015-08-10
Fixed #696 (Support for CBLReplication.pendingDocumentIds)

### LINKED ISSUE #696: Support for CBLReplication.pendingDocumentIds
This feature has been implemented in iOS and was recently fixed to return the result even when the replication is not running. It's returned the document ids that haven't been pushed to the server.

Issue: https://github.com/couchbase/couchbase-lite-ios/issues/807

hideki:
- Solution: https://github.com/couchbase/couchbase-lite-ios/pull/819

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -75,8 +75,6 @@ public class Database implements StoreDelegate {
     // Default value for maxRevTreeDepth, the max rev depth to preserve in a prune operation
     public static int DEFAULT_MAX_REVS = 20;
 
-    private static String kLocalCheckpointDocId = "CBL_LocalCheckpoint";
-
 
     private Store store = null;
     private String path;
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -1220,20 +1220,6 @@ public class Database implements StoreDelegate {
     }
 
     @InterfaceAudience.Private
-    public Map<String, Object> getLocalCheckpointDocument() {
-        return this.getExistingLocalDocument(kLocalCheckpointDocId);
-    }
-
-    @InterfaceAudience.Private
-    public String getLastSequenceForReplicator(String checkpointID, String fallbackCheckpointID) {
-        String lastSequence = lastSequenceWithCheckpointId(checkpointID);
-        if (lastSequence == null && fallbackCheckpointID != null) {
-            lastSequence = lastSequenceWithCheckpointId(fallbackCheckpointID);
-        }
-        return lastSequence;
-    }
-
-    @InterfaceAudience.Private
     public RevisionList unpushedRevisionsSince(String sequence,
                                                ReplicationFilter filter,
                                                Map<String, Object> filterParams) {
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -75,6 +75,8 @@ public class Database implements StoreDelegate {
     // Default value for maxRevTreeDepth, the max rev depth to preserve in a prune operation
     public static int DEFAULT_MAX_REVS = 20;
 
+    private static String kLocalCheckpointDocId = "CBL_LocalCheckpoint";
+
 
     private Store store = null;
     private String path;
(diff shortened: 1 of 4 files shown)


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -406,20 +404,83 @@ public int getChangesCount() {
...
+            latch.await();
+        } catch (InterruptedException e) {
+            Log.e(Log.TAG_SYNC, "InterruptedException", e);
+        }
+
+        return pendingDocIDs;
+    }
+
+    public boolean isDocumentPending(Document doc){
+        return doc != null && getPendingDocumentIDs().contains(doc.getId());
```
pasin: I think we also need to check that getPendingDocumentIDs() returns null or not.

hideki:
@pasin,
I updated PR. Please review this again?

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -75,6 +75,8 @@
     // Default value for maxRevTreeDepth, the max rev depth to preserve in a prune operation
     public static int DEFAULT_MAX_REVS = 20;
 
+    private static String kLocalCheckpointDocId = "CBL_LocalCheckpoint";
```
pasin: I think this is non-used and can be removed.

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -406,20 +404,76 @@ public int getChangesCount() {
...
+    /**
+     * Following two methods for temporary methods instead of CBL_ReplicatorSettings implementation.
+     */
+    protected String remoteCheckpointDocID() {
         return replicationInternal.remoteCheckpointDocID();
     }
 
+    public Set<String> getPendingDocumentIDs(){
+        if(isPull() || (isRunning() && pendingDocIDs != null))
+            return pendingDocIDs;
```
pasin: I couldn't find where the pendingDocIDs gets reset inside the Replicator class. It should get reset when the replicator changes its status. Here is where iOS does it.

https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/API/CBLReplication.m#L349

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -406,20 +404,76 @@ public int getChangesCount() {
...
+                    ReplicationFilter filter = null;
+                    if (filterName != null) {
+                        filter = db.getFilter(filterName);
+                    }
+
+                    String lastSequence = replicationInternal.lastSequence;
+                    if (lastSequence == null)
+                        lastSequence = db.lastSequenceWithCheckpointId(remoteCheckpointDocID());
+                    RevisionList revs = db.unpushedRevisionsSince(lastSequence, filter, filterParams);
+                    Log.e(Log.TAG_SYNC, revs.toString());
```
pasin: I saw the line below checking `revs != null`. If revs is null, this will be NPE.

hideki: It is for debugging. I should remove before commit.

hideki:
Hi @pasin,
I applied your feedbacks. Gould you please review again?

src/main/java/com/couchbase/lite/replicator/Replication.java
pasin: As the `pendingDocIDs` is reset on the work executor thread, to avoid race condition, I think accessing `pendingDocIDs` variable here should be in a synchronized block.

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -369,6 +377,15 @@ public ReplicationStatus getStatus() {
      */
     @Override
     public void changed(ChangeEvent event) {
+
+        // forget cached IDs (Should be executed in workExecutor)
+        db.runAsync(new AsyncTask() {
+            @Override
+            public void run(Database database) {
+                pendingDocIDs = null;
```
pasin: I think resetting the `pendingDocIDs` variable here needs a synchronized block as the variable is also accessed on the main thread in the getPendingDocumentIDs() method below.

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -184,6 +186,15 @@ public void start() {
             }
         }
 
+        // forget cached IDs (Should be executed in workExecutor)
+        db.runAsync(new AsyncTask() {
+            @Override
+            public void run(Database database) {
+                pendingDocIDs = null;
```
pasin: I think resetting the pendingDocIDs variable here needs a synchronized block as the variable is also accessed on the main thread in the getPendingDocumentIDs() method below.

src/main/java/com/couchbase/lite/replicator/Replication.java
pasin: What thread does the changed() method get executed?

hideki: If SYNC mode, whatever original thread is.
If ASYNC mode, it is called from workExecutor (single thread).

However, `changeListenerNotifyStyle` is always `ChangeListenerNotifyStyle.ASYNC`. 

```java
f (changeListenerNotifyStyle == ChangeListenerNotifyStyle.SYNC) {
            for (ChangeListener changeListener : changeListeners) {
                try {
                    changeListener.changed(changeEvent);
                } catch (Exception e) {
                    e.printStackTrace();
                    Log.e(Log.TAG_SYNC, "Exception notifying replication listener: %s", e);
                }
            }
        } else {
            // ASYNC
            synchronized (workExecutor) {
                if (!workExecutor.isShutdown()) {
                    workExecutor.submit(new Runnable() {
                        @Override
                        public void run() {
                            try {
                                for (ChangeListener changeListener : changeListeners) {
                                    changeListener.changed(changeEvent);
                                }
                            } catch (Exception e) {
                                Log.e(Log.TAG_SYNC, "Exception notifying replication listener: %s", e);
                                throw new RuntimeException(e);
                            }
                        }
                    });
                }
            }
        }
```

hideki:
Hi @pasin,
I applied your feedbacks. Gould you please review again?

hideki:
Hi @pasin,
I applied your feedbacks again. Gould you please review again?

Issue #710, opened 2015-08-21
Encrypt file attachments


zgramana:

pasin:
I have worked on this together with the database encryption.
https://github.com/couchbase/couchbase-lite-java-core/pull/730

PR #730, opened 2015-09-10
Fixed #702 and #710 Implement database and attachment encryption

### LINKED ISSUE #702: Add support for SQLCipher

pasin:
- Implemented database encryption with android-database-sqlcipher
- Implmented attachment encryption
- Implemented SymmetricKey class to facilitate key generation and encryption/descryption
- Added enableStorageEncryption option to ManagerOption to enable storage encryption

#702 #710

### DIFF
--- a/src/main/java/com/couchbase/lite/Attachment.java
+++ b/src/main/java/com/couchbase/lite/Attachment.java
@@ -19,6 +19,7 @@ package com.couchbase.lite;
 
 import com.couchbase.lite.internal.AttachmentInternal;
 import com.couchbase.lite.internal.InterfaceAudience;
+import com.couchbase.lite.support.security.SymmetricKeyException;
 import com.couchbase.lite.util.Log;
 
 import java.io.ByteArrayInputStream;
@@ -188,7 +189,6 @@ public final class Attachment {
     protected static Map<String, Object> installAttachmentBodies(Map<String, Object> attachments,
                                                                  Database database)
             throws CouchbaseLiteException {
-
         Map<String, Object> updatedAttachments = new HashMap<String, Object>();
         for (String name : attachments.keySet()) {
             Object value = attachments.get(name);
@@ -202,7 +202,7 @@ public final class Attachment {
                     BlobStoreWriter writer;
                     try {
                         writer = blobStoreWriterForBody(body, database);
-                    } catch (IOException e) {
+                    } catch (Exception e) {
                         throw new CouchbaseLiteException(e.getMessage(), Status.ATTACHMENT_ERROR);
                     }
                     metadataMutable.put("length", writer.getLength());
@@ -222,12 +222,15 @@ public final class Attachment {
 
     @InterfaceAudience.Private
     protected static BlobStoreWriter blobStoreWriterForBody(InputStream body, Database database)
-            throws IOException {
+            throws IOException, SymmetricKeyException {
         BlobStoreWriter writer = database.getAttachmentWriter();
         try {
-            writer.read(body);
+            writer.appendInputStream(body);
             writer.finish();
         } catch (IOException e) {
+            writer.cancel();
+            throw e;
+        } catch (SymmetricKeyException e) {
             writer.cancel();
             throw e;
         }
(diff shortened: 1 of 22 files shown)


src/main/java/com/couchbase/lite/Attachment.java
```diff
@@ -202,7 +202,7 @@ protected InputStream getBodyIfNew() {
                     BlobStoreWriter writer;
                     try {
                         writer = blobStoreWriterForBody(body, database);
-                    } catch (IOException e) {
+                    } catch (Exception e) {
                         throw new CouchbaseLiteException(e.getMessage(), Status.ATTACHMENT_ERROR);
```
hideki: Don't we need to treat SymmetricKeyException and IOException as different error? or Is it ok to treat both exception as just attachment error?

src/main/java/com/couchbase/lite/BlobStore.java
```diff
@@ -41,28 +45,36 @@
...
-                Log.w(Log.TAG_DATABASE, "Unable to make directory: %s", directory);
-        }
-        if (!directory.isDirectory()) {
-            throw new IllegalStateException(String.format("Unable to create directory for: %s",
-                    directory));
+        File directory = new File(path);
+        if (directory.exists()) {
+            if (!directory.isDirectory())
+                throw new CouchbaseLiteException("BlobStore: Blobstore is not a directory", Status.ATTACHMENT_ERROR);
+            verifyExistingStore();
```
hideki: Is there scenario that directory exists but marker file does not exist??

src/main/java/com/couchbase/lite/Manager.java
```diff
@@ -342,8 +400,16 @@ public boolean replaceDatabase(String databaseName, String databaseDir) {
...
         }
 
-        if(!db.open()){
-            Log.w(Database.TAG, "Failed to open database");
+        boolean isOpen = false;
+        CouchbaseLiteException error = null;
+        try {
+            isOpen = db.open();
+        } catch (CouchbaseLiteException e) {
+            error = e;
```
hideki: Why not return false from catch block?

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -275,7 +275,14 @@ private Status openDB() {
         if (!db.exists()) {
             return new Status(Status.NOT_FOUND);
         }
-        if (!db.open()) {
+
+        boolean isOpen = false;
+        try {
+            isOpen = db.open();
```
hideki: [Discussion Item] Should Database.open() throw CouchbaseLiteException if it fails to open()?

src/main/java/com/couchbase/lite/Attachment.java
pasin: I would think this is enough. iOS doesn't have anything special for this as well.

src/main/java/com/couchbase/lite/BlobStore.java
pasin: In general cases, there shouldn't be the case. Jens added this to fix a bug that the iOS doesn't really encrypting the attachments.

VerifyingExistingStore does 2 things for ensuring things work as expected and if not (for some reasons such as bugs), instead of failing silently, an error will be thrown:
1. Verifying the marker file exists otherwise error should be thrown.
2. Verifying an encryption algorithm.

src/main/java/com/couchbase/lite/router/Router.java
pasin: I remembered asking you about this before. Currently in the master branch, no exception thrown (only true / false) so no ways to know the cause of error. I believe we should throw an exception. iOS always has `out error` parameter to check what the error is.

While I'm working for SQLiteDatabase, Android always throws SQLiteDatabaseException or one of its subclass exception BUT they made them as a runtime exception. We could explore this option as well.

pasin: Another option would be instead of return true / false, returning as a status code.

pasin: We should discuss more about this and make it as a standard for cbl java/android.

src/main/java/com/couchbase/lite/Manager.java
pasin: Because isOpen could be false without throwing an exception (I haven't fixed to throw an exception for all error cases).

PR #734, opened 2015-09-11
Fixed #729

### LINKED ISSUE #729: ApiTest:testViewWithLinkedDocs hangs when running unit tests with couchbase-lite-java
It seems like this code never exists from the loop.

https://github.com/couchbase/couchbase-lite-java-core/blob/master/src/main/java/com/couchbase/lite/Database.java#L1844-L1853

hideki:
- SQLiteViewStore did not handle create attribute properly.  Fixed this.

Equivalent iOS code
https://github.com/couchbase/couchbase-lite-ios/blob/f38ef1e54338e10d299eb1ccbc6f4c4bd442eb5a/Source/CBL_SQLiteViewStorage.m#L56-L57


pasin:
Tested it and the fixed solved the issue.

PR #759, opened 2015-09-25
Fixed #664 - Creating authenticator with empty password throws ArrayI…

### LINKED ISSUE #664: Creating authenticator with empty password throws ArrayIndexOutOfBoundsException
### Given
```
com.couchbase.lite.auth.Authenticator auth = AuthenticatorFactory.createBasicAuthenticator("GUEST", "");
```
### Expected
Empty password sent during HTTP challenge/response.

### Observed
```
java.lang.ArrayIndexOutOfBoundsException: 1
at com.couchbase.lite.support.RemoteRequest.preemptivelySetAuthCredentials(RemoteRequest.java:281)
```
### Note
There's an unchecked array index access after a string split on [line 271](https://github.com/couchbase/couchbase-lite-java-core/blob/733954834d71d7b15426f4c685c8ff002461bbb1/src/main/java/com/couchbase/lite/support/RemoteRequest.java#L271).

### Workaround

In the case of Sync Gateway's GUEST user, no authenticator is used at all. In fact, the GUEST user is the automatic user context for all anonymous, unauthenticated users.

hideki:
…ndexOutOfBoundsException

In case userInfo is `username:`, userInfoElements is `['username']`. So just added line to check size of userInfoElements array.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/RemoteRequest.java
+++ b/src/main/java/com/couchbase/lite/support/RemoteRequest.java
@@ -267,10 +267,11 @@ public class RemoteRequest implements Runnable {
         if (userInfo != null) {
             if (userInfo.contains(":") && !userInfo.trim().equals(":")) {
                 String[] userInfoElements = userInfo.split(":");
-                String username = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[0]): userInfoElements[0];
-                String password = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[1]): userInfoElements[1];
+                String username = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[0]) : userInfoElements[0];
+                String password = "";
+                if(userInfoElements.length >= 2)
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/support/RemoteRequest.java
```diff
@@ -267,10 +267,11 @@ protected void preemptivelySetAuthCredentials(HttpClient httpClient) {
         if (userInfo != null) {
             if (userInfo.contains(":") && !userInfo.trim().equals(":")) {
                 String[] userInfoElements = userInfo.split(":");
-                String username = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[0]): userInfoElements[0];
-                String password = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[1]): userInfoElements[1];
+                String username = isUrlBasedUserInfo ? URIUtils.decode(userInfoElements[0]) : userInfoElements[0];
+                String password = "";
```
pasin: I believe @hideki already checked if it should be an empty string instead of null value.

PR #774, opened 2015-10-02
Fixed #719 #720 Support SQLite database and attachment rekey

### LINKED ISSUE #719: Support SQLite encryption rekey feature
iOS References:
https://github.com/couchbase/couchbase-lite-ios/commit/ad82b1b94887df639034baf3dfaa1812508fd60a
https://github.com/couchbase/couchbase-lite-ios/commit/ddf987bf267c6752b5c44a472d1556f27bcd9276
https://github.com/couchbase/couchbase-lite-ios/commit/e371f531085f1f978a60c814cbadc738312c705c

pasin:
- Implemented SQLite database and attachment rekey
- API Change:
[Added] Database.changeEncryptionKey(final Object newKeyOrPassword)
[Added] Manager.registerEncryptionKey(Object keyOrPassword, String databaseName)
[Removed] Manager.registerEncryptionKey(String password, String databaseName)
[Removed] Manager.registerEncryptionKey(byte[] keys, String databaseName)
[Added] Context.getTempDir()

#719 #720

### DIFF
--- a/src/main/java/com/couchbase/lite/BlobStore.java
+++ b/src/main/java/com/couchbase/lite/BlobStore.java
@@ -17,6 +17,10 @@
 
 package com.couchbase.lite;
 
+import com.couchbase.lite.support.FileDirUtils;
+import com.couchbase.lite.support.action.Action;
+import com.couchbase.lite.support.action.ActionBlock;
+import com.couchbase.lite.support.action.ActionException;
 import com.couchbase.lite.support.security.SymmetricKey;
 import com.couchbase.lite.support.security.SymmetricKeyException;
 import com.couchbase.lite.util.Log;
@@ -50,15 +54,18 @@ public class BlobStore {
     public static final String ENCRYPTION_ALGORITHM = "AES";
     public static final String ENCRYPTION_MARKER_FILENAME = "_encryption";
 
+    private Context context;
     private String path;
     private SymmetricKey encryptionKey;
+    private BlobStore tempStore;
 
-    public BlobStore(String path, SymmetricKey encryptionKey) throws CouchbaseLiteException {
-        this(path, encryptionKey, false);
+    public BlobStore(Context context, String path, SymmetricKey encryptionKey) throws CouchbaseLiteException {
+        this(context, path, encryptionKey, false);
     }
 
-    public BlobStore(String path, SymmetricKey encryptionKey, boolean autoMigrate)
+    public BlobStore(Context context, String path, SymmetricKey encryptionKey, boolean autoMigrate)
             throws CouchbaseLiteException {
+        this.context = context;
         this.path = path;
         this.encryptionKey = encryptionKey;
 
@@ -106,10 +113,17 @@ public class BlobStore {
             }
         } else if (!isMarkerExists) {
             // No "_encryption" file was found, so on-disk store isn't encrypted:
-            if (encryptionKey != null) {
-                // This shouldn't be happening:
-                Log.e(Log.TAG_DATABASE, "BlobStore: Blob store was corrupted. Encryption marker file was not found");
-                throw new CouchbaseLiteException(Status.CORRUPT_ERROR);
+            SymmetricKey encKey = encryptionKey;
+            if (encKey != null) {
+                // In general, this case shouldn't happen but just in case:
+                Log.i(Log.TAG_DATABASE, "BlobStore: BlobStore should be encrypted; do it now...");
+                encryptionKey = null;
+                try {
+                    changeEncryptionKey(encKey);
+                } catch (ActionException e) {
+                    new CouchbaseLiteException("Cannot change the attachment encryption key", e,
+                            Status.ATTACHMENT_ERROR);
+                }
             }
         }
     }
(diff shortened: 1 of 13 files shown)


src/main/java/com/couchbase/lite/BlobStore.java
```diff
@@ -138,6 +152,157 @@ private void markEncrypted(boolean encrypted) throws CouchbaseLiteException {
...
+            return action;
+        }
+
+        // Create a new directory for the new blob store. Have to do this now, before starting the
+        // action, because farther down we create an action to move it...
+        File tempDir = new File (context.getTempDir(), Misc.CreateUUID());
+        final File tempStoreDir = tempDir.mkdirs() ? tempDir : null;
+        // Mark delete on exit:
+        if (tempStoreDir != null)
+            tempStoreDir.deleteOnExit();
```
hideki: http://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()
`deleteOnExit()` works with normal JVM termination. Abnormal case might not work as expected.
And, I am not sure Android's JVM life cycle. Does this line work as expected with Android?

src/main/java/com/couchbase/lite/support/action/Action.java
```diff
@@ -0,0 +1,342 @@
...
+        return failedStep;
+    }
+
+    /**
+     * Adds an action as a step of thid one.
+     * @param action
+     */
+    public void add(final AtomicAction action) {
+        if (action instanceof Action) {
+            Action a = (Action)action;
```
hideki: Does not need to check number of perfroms, backouts, and cleanUps before adding them into list? I assume they should be same size, otherwise it might cause problem.

src/main/java/com/couchbase/lite/support/action/Action.java
```diff
@@ -0,0 +1,342 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void backout() throws ActionException {
+        if (nextStep == 0)
+            throw new IllegalStateException("Actions have not been run");
+
+        while(nextStep-- > 0) {
```
hideki: I guess backout for current step is not executed. Is it expected?

If error occurs with current step, how does it recover it?

src/main/java/com/couchbase/lite/support/action/Action.java
```diff
@@ -0,0 +1,342 @@
...
+     * If the actions succeeded, cleans them up in reverse order.
+     * The `lastError` property is set to the exception thrown by the failed perform block.
+     * The `failedStep` property is set to the index of the failed perform block.
+     * @throws ActionException
+     */
+    public void run() throws ActionException {
+        try {
+            perform();
+            try {
+                cleanup(); // Ignore exception
```
hideki: In case error occurs, `cleanup()` is not called. Is it expected behavior?

src/main/java/com/couchbase/lite/support/action/Action.java
```diff
@@ -0,0 +1,342 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * An AtomicAction implementation to provide an ability to perform actions,
+ * back them out if there is an error, and clean up.
+ */
+public class Action implements AtomicAction {
```
hideki: Did you write Test for Action? I recommend to write test for normal case, backout case.

src/main/java/com/couchbase/lite/support/action/Action.java
```diff
@@ -0,0 +1,342 @@
...
+
+    /** File-based actions: **/
+
+    /**
+     * Deletes the file/directory at the given path, if it exists.
+     * @param path Path to the file or directory to be deleted
+     * @param tempDir Path to a tem directory used for backing out the deleted file
+     * @return A delete file Action
+     */
+    public static Action deleteFile(final String path, final String tempDir) {
```
hideki: `deleteFile()` and `moveFileToEmptyPath()` methods are an utility method. I recommend to extract these method from Action.java, and create new Java Class.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -351,6 +354,125 @@ private void decrypt(SymmetricKey encryptionKey) throws CouchbaseLiteException {
...
+    // ENCRYPTABLE STORE
+    ///////////////////////////////////////////////////////////////////////////
+
+    @Override
+    public void setEncryptionKey(SymmetricKey key) {
+        encryptionKey = key;
+    }
+
+    @Override
+    public Action actionToChangeEncryptionKey(final SymmetricKey newKey) {
```
hideki: Discussion Item: It seems actionToChangeEncryptionKey() can work only with SQLCipher. Not with default SQLite. Should it be implemented outside of SQLiteStore?

I am not against current structure. Just I'd like to have cleaner isolation if possible.

src/main/java/com/couchbase/lite/BlobStore.java
pasin: This is just a good to have trick that I added so the temp file can be more guaranteed to be deleted. We already cleaned up the file in the cleanup block.

I understand that this feature doesn't work for android but it's ok.

src/main/java/com/couchbase/lite/support/action/Action.java
pasin: There is no need to check because as you said no methods to add an individual action.

src/main/java/com/couchbase/lite/support/action/Action.java
pasin: Yes based on the iOS implementation.

src/main/java/com/couchbase/lite/support/action/Action.java
pasin: Backing out happen when the next perform step fails. When a perform action failed, it usually means that the thing doesn't happen and the perform action itself should make sure that it has handle that properly.

Anyway, here are the full definitions:
```
/**
* Adds an action as a step of this one. The action has three components, each optional.
* @param perform A block that tries to perform the action, or returns an error if it fails.
*                (If the block fails, it should clean up; the backOut will _not_ be called!)
* @param backout A block that undoes the effect of the action; it will be called if a _later_
*                action fails, so that the system can be returned to the initial state.
* @param cleanup A block that performs any necessary cleanup after all actions have been
*                performed (e.g. deleting a temporary file.)
*/
```

src/main/java/com/couchbase/lite/support/action/Action.java
pasin: I would say let make them here for now to be matched with the iOS implementation. If in the future it starts to get bigger and hard to maintain, we could move.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
pasin: The method is a part of the EncryptableStore interface so basically it comes together with setEncryptionKey() method. Currently the SQLiteStore itself also implements the encryption logic and it uses storageEngine.supportEncryption() method to determine if the encryption feature is available.

In my opinion: 
1. It's ok to be here unless we decide to extract the encryption feature out of a regular SQLiteStore (e.g. create a new SQLCipherStore). However making SQLCipherStore separately would need to have a factory object to determine what store to be created (which I think we already have that for creating SQLiteStore and ForestDBStore). 

2. If possible, let wait until the ForestDBStore encryption get implemented because they should get implemented in the same or similar structure. 

3. I don't like to say this but for code maintenance to match with iOS, sometimes it's useful to stay close to the iOS implementation as much as possible. Anyway, this implementation is currently matched with the iOS :)

src/main/java/com/couchbase/lite/support/action/Action.java
pasin: Thanks! I have ported the test on my personal mac but forgot to move it. Will do it.

pasin: I have brought it in.

src/main/java/com/couchbase/lite/BlobStore.java
pasin: BTW, I just added a small comment to the code and checked it in.

src/main/java/com/couchbase/lite/support/action/Action.java
hideki:
```
* @param perform A block that tries to perform the action, or returns an error if it fails.
*                (If the block fails, it should clean up; the backOut will _not_ be called!)
```
perform should clean up itself if error occurs. Thanks.

PR #782, opened 2015-10-12
Fixed #742 Refactor Store.open and Database.open method signature

### LINKED ISSUE #742: Make Store.open() method return void and throw exception when having an error
Per discussion with @hideki, we should make this method to return void and throw exception when there is an error (instead of mixing returning true/false with throwing an exception).

pasin:
Made Store.open and Database.open to return void and throw CouchbaseLiteException.

#742

### DIFF
--- a/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
+++ b/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
@@ -86,7 +86,7 @@ final public class DatabaseUpgrade {
             try {
                 db.open();
             } catch (CouchbaseLiteException e) {
-                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", db.toString(), e);
+                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", e, db.toString());
                 return false;
             }
 
--- a/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
+++ b/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
@@ -83,15 +83,10 @@ final public class DatabaseUpgrade {
             }
 
             // Open destination database:
-            boolean isOpen = false;
-            Throwable error = null;
             try {
-                isOpen = db.open();
+                db.open();
             } catch (CouchbaseLiteException e) {
-                error = e;
-            }
-            if (!isOpen) {
-                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", db.toString(), error);
+                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", db.toString(), e);
                 return false;
             }
 
--- a/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
@@ -753,7 +753,7 @@ abstract class ReplicationInternal implements BlockingQueueListener {
                                 isOpen = true;
                             }
                         } catch (CouchbaseLiteException ex) {
-                            Log.w(Log.TAG_SYNC, "%s: Cannot open the database", this, ex);
+                            Log.w(Log.TAG_SYNC, "%s: Cannot open the database", ex, this);
                         }
 
                         if (isOpen) {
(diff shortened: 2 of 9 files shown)


src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -197,77 +210,84 @@ public synchronized boolean open() throws CouchbaseLiteException {
...
                 // monotonically increasing, never reused. See <http://www.sqlite.org/autoinc.html>)
-
                 if (!isNew) {
-                    Log.w(TAG, "Database version (%d) is older than I know how to work with",
-                            dbVersion);
-                    storageEngine.close();
-                    return false;
+                    String message = "Database version " + dbVersion +
+                            " is older than I know how to work with";
+                    Log.e(TAG, message, dbVersion);
```
hideki: In format, no %d. But method is given dbVersion as third parameter. I am not sure what kind of problem this might cause.

src/main/java/com/couchbase/lite/DatabaseUpgrade.java
```diff
@@ -83,15 +83,10 @@ protected boolean importData() {
...
-            Throwable error = null;
             try {
-                isOpen = db.open();
+                db.open();
             } catch (CouchbaseLiteException e) {
-                error = e;
-            }
-            if (!isOpen) {
-                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", db.toString(), error);
+                Log.e(TAG, "Upgrade failed:  Couldn't open new db: %s", db.toString(), e);
```
hideki: Original code was wrong. Third parameter should be Throwable.
```java
    public static void e(String tag, String formatString, Throwable tr, Object... args) {
        if (logger != null && isLoggingEnabled(tag, ERROR)) {
            try {
                logger.e(tag, String.format(formatString, args), tr);
            } catch (Exception e) {
                logger.e(tag, String.format("Unable to format log: %s", formatString), e);
            }
        }
    }
```

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -748,7 +748,10 @@ public void onCompletion(HttpResponse httpResponse, Object result, Throwable e)
                         remoteCheckpoint = body;
                         boolean isOpen = false;
                         try {
-                            isOpen = (db != null && db.open());
+                            if (db != null) {
+                                db.open();
+                                isOpen = true;
+                            }
                         } catch (CouchbaseLiteException ex) {
                             Log.w(Log.TAG_SYNC, "%s: Cannot open the database", this, ex);
```
hideki: Third parameter should be Throwable.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -197,77 +210,84 @@ public synchronized boolean open() throws CouchbaseLiteException {
 
         // Incompatible version changes increment the hundreds' place:
         if (dbVersion >= 200) {
-            Log.e(TAG, "Database: Database version (%d) is newer than I know how to work with",
-                    dbVersion);
             storageEngine.close();
-            return false;
+            String message = "Database version " + dbVersion +
+                    " is newer than I know how to work with";
+            Log.e(TAG, message, dbVersion);
```
hideki: No %d in format. But third parameter is given.

src/main/java/com/couchbase/lite/DatabaseUpgrade.java
pasin: I have noticed that DatabaseUpgrade.java has several Log method calls that I'm not sure about the original attention. For example,

```
Log.w(TAG, "Couldn't import local doc '%s': '%s'", docID, e);
```
There are several like the example above. All of these should be revise.

pasin:
Thanks for the review. I have pushed up the changes. Please re-review.

src/main/java/com/couchbase/lite/DatabaseUpgrade.java
hideki: I guess `e` intents `e.toString()`.

pasin: Why don't just put e in the 3rd parameter? Anyway I think this is a bit off topic of this PR.

PR #784, opened 2015-10-12
Fixed #781 added code to handle the case of large file

### LINKED ISSUE #781: Review BlobStore.getBytesFromFile if the file is large.
Continue from https://github.com/couchbase/couchbase-lite-java-core/pull/779.

1. Verify which parts of code is using this method.
2. Verify if the method handle the case of large file.

pasin:
I have checked the implementation of Java7's Files.readAllBytes() and the implementation throws OOM exception if the size is larger than Interger.MAX so I did the same thing. Also per @hideki suggested, I also added a check for null value when creating a byte array object to throw OOM exception if it's null.

Reference: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7-b147/java/nio/file/Files.java#Files.readAllBytes%28java.nio.file.Path%29

#781

### DIFF
--- a/src/main/java/com/couchbase/lite/BlobStore.java
+++ b/src/main/java/com/couchbase/lite/BlobStore.java
@@ -558,13 +558,17 @@ public class BlobStore {
         try {
             is = new FileInputStream(file);
 
-            // Get the size of the file
+            // Get the size of the file:
             long length = file.length();
+            if (length > (long)Integer.MAX_VALUE)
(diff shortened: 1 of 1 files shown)


hideki:
This OutOfMemory error could be recoverable. Please catch the Error in caller side to handle error appropriately.

pasin:
OK, I handled the OOM on the caller side as suggested. Thanks!

Issue #805, opened 2015-11-05
Implement PBKDF2 Key Derivation with libcrypto


pasin:
For binary size optimization, this requires to move the key derivation JNI into the storage engine level.

PR #806, opened 2015-11-05
Issue #805 Implement key derivation with libcrypto

pasin:
- Removed SymmetricKey JNI with libcrypto

- Brought back the single build.gradle (no separated java and android)

- Implemented Key Derivation at the Store and Storage Level.

- Refactor SymmetricKey to accept only raw key (byte array), no password string. Database will derive the raw key from the password when creating a SymmetricKey object. However when doing this, the storage engine needs to be created prior openning the store in Store.open.

#805


src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -166,17 +167,12 @@ public boolean databaseExists(String directory) {
...
-                manager.getContext().getSQLiteStorageEngineFactory();
-        storageEngine = factory.createStorageEngine(manager.isEnableStorageEncryption());
-
         // Try to open the storage engine and stop if we fail:
-        if (storageEngine == null) {
-            String message = "Unable to create a storage engine, fatal error";
-            Log.e(TAG, message);
-            throw new IllegalStateException(message);
-        }
+        if (storageEngine == null)
```
hideki: Constructor initializes storageEngine. I think this is redundant.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -145,7 +146,7 @@ public SQLiteStore(String directory, Manager manager, StoreDelegate delegate) {
         }
         this.path = new File(directory, kDBFilename).getPath();
         this.manager = manager;
-        this.storageEngine = null;
+        this.storageEngine = createStorageEngine();
```
hideki: close() method might set null for storageEngine. So open() method needs to check if storageEngine is null.
I think storageEngine initialization is not necessary in Constructor.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -490,6 +499,22 @@ public void execute() throws ActionException {
         return action;
     }
 
+    @Override
+    public byte[] derivePBKDF2SHA256Key(String password, byte[] salt, int rounds)
+            throws CouchbaseLiteException {
+        if (!storageEngine.supportEncryption()) {
```
hideki: null check might be required here.

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -74,6 +74,10 @@
     // When this many changes pile up in _changesToNotify, start removing their bodies to save RAM
     private static final int MANY_CHANGES_TO_NOTIFY = 5000;
 
+    private static final String DEFAULT_PBKDF2_KEY_SALT = "Salty McNaCl";
```
hideki: What is "Salty McNaCl"?

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1076,8 +1077,10 @@ public synchronized void open() throws CouchbaseLiteException {
             store = new SQLiteStore(path, manager, this);
 
         // Set encryption key:
-        SymmetricKey encryptionKey = manager.getEncryptionKeys().get(name);
+        SymmetricKey encryptionKey = null;
         if (store instanceof EncryptableStore) {
+            Object keyOrPassword = manager.getEncryptionKeys().get(name);
+            encryptionKey = createSymmetricKey(keyOrPassword);
             ((EncryptableStore)store).setEncryptionKey(encryptionKey);
```
hideki: Question: Does `setEncryptionKey(..)` allow null as its parameter?

src/main/java/com/couchbase/lite/store/SQLiteStore.java
pasin: It's required because the storage engine now is needed for doing key derivation which will happen before open() method is called.

https://github.com/couchbase/couchbase-lite-java-core/blob/feature/issue_805_key_derivation/src/main/java/com/couchbase/lite/Database.java#L1080-L1088

src/main/java/com/couchbase/lite/store/SQLiteStore.java
pasin: Agree. Will handle that.

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1142,6 +1145,39 @@ public synchronized void open() throws CouchbaseLiteException {
...
+        if (keyOrPassword == null)
+            return null;
+
+        byte[] rawKey = null;
+        if (keyOrPassword instanceof String) {
+            rawKey = ((EncryptableStore) store).derivePBKDF2SHA256Key(
+                    (String)keyOrPassword,
+                    DEFAULT_PBKDF2_KEY_SALT.getBytes(),
+                    DEFAULT_PBKDF2_KEY_ROUNDS);
+        } else {
```
hideki: Personally I prefer
```
if(xxx instanceof String){
}
else if(xxx instanceof byte[]){
}
else{
}
```

src/main/java/com/couchbase/lite/store/SQLiteStore.java
pasin: Another way to handle this is checking whether the storage is null or not in derivePBKDF2SHA256Key(String password, byte[] salt, int rounds) method below. If it's null, create a new one. I think this is a better approach. What do you think?

src/main/java/com/couchbase/lite/Database.java
pasin: It's just a default salt used for deriving PBDKF2. The is also the same as the iOS implementation. I think NaCl is Sodium Chloride. I don't know what is Mc.

src/main/java/com/couchbase/lite/Database.java
pasin: Yes.

src/main/java/com/couchbase/lite/Database.java
pasin: Do you suggested to changer? Personally I prefer the default convention from Android Studio so we will not forget to do it differently.

pasin: I may not understand your point here. Are you suggesting the convention with else if under the } or something else?

pasin: I may got your point now. Will change that.

src/main/java/com/couchbase/lite/store/SQLiteStore.java
hideki: I prefer your second option. Because storageEgnine could be null when `derivePBKDF2SHA256Key()` is called.

pasin:
@hideki , I made changes per your review. Could you please re-review? Thanks!

PR #815, opened 2015-11-12
Fixed https://github.com/couchbase/couchbase-lite-android/issues/683

hideki:
DocumentChange of Database.ChangeEvent might not contain revision body. In case of include_docs=true, needs to load revision body

Test is done by Sync Gateway, https://github.com/couchbase/couchbase-lite-android-liteserv and Curl command [Makefile](https://gist.github.com/hideki/823c1c02157a3f9f1b2f)

### DIFF
--- a/src/main/java/com/couchbase/lite/router/Router.java
+++ b/src/main/java/com/couchbase/lite/router/Router.java
@@ -1436,13 +1436,25 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
 
             RevisionInternal rev = change.getAddedRevision();
 
+            // https://github.com/couchbase/couchbase-lite-android/issues/683
+            // DocumentChange of Database.ChangeEvent might not contain revision body.
+            // In case of include_docs=true, needs to load revision body
+            if (changesIncludesDocs && (rev.getBody() == null || !rev.getBody().hasContent())) {
+                EnumSet<TDContentOptions> options = getContentOptions();
+                try {
+                    rev = db.loadRevisionBody(rev, options);
+                } catch (CouchbaseLiteException e) {
+                    Log.w(Log.TAG_ROUTER, "Error loading revision body", e);
+                }
+            }
+
             final boolean allowRevision = event.getSource().runFilter(changesFilter, changesFilterParams, rev);
             if (!allowRevision) {
                 return;
             }
 
             if (longpoll) {
-                Log.w(Log.TAG_ROUTER, "Router: Sending longpoll response");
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1436,13 +1436,25 @@ public void changed(Database.ChangeEvent event) {
...
+            // In case of include_docs=true, needs to load revision body
+            if (changesIncludesDocs && (rev.getBody() == null || !rev.getBody().hasContent())) {
+                EnumSet<TDContentOptions> options = getContentOptions();
+                try {
+                    rev = db.loadRevisionBody(rev, options);
+                } catch (CouchbaseLiteException e) {
+                    Log.w(Log.TAG_ROUTER, "Error loading revision body", e);
+                }
+            }
+
```
pasin: The code here is out-of-sync with the iOS implementation. There are details about includesConflicts & includesDocs and the winning revision that should be ported here.

iOS Implementation: https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLRouter%2BChanges.m#L182-L201

[API docs] (
http://developer.couchbase.com/documentation/mobile/1.1.0/develop/references/couchbase-lite/rest-api/database/get-changes/index.html) for more info about includeConflicts and includesDocs.

hideki: @pasin,
Question about iOS implementation: In case winningRevID and rev.revID is same, iOS implementation does not load revision body. iOS `CBLDatabaseChange` always has Revision body? 
CBL Android/Java does not guarantee.

pasin: I understand that the `rev` already contains the properties except deleted revision.

https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBL_SQLiteStorage.m#L1646-L1650

@snej could you help confirming this?

snej: Yes, I think it includes the properties. It's cheaper than having to read them from the database again.

hideki: Hi @pasin and @snej,
Thank you very much reviewing code during night. I believe this issue caused by releasing body field aggressively to release memory during pull replication to reduce peek memory usage. It seems that fixing this problem in Router.java is not right approach. I will cancel this PR.
About conflicted document, I will create another ticket to track.
Again Thanks a lot!!

hideki:
Fixing the issue in Router.java is not right approach. Closed this PR.

PR #817, opened 2015-11-12
Fixed https://github.com/couchbase/couchbase-lite-android/issues/683

hideki:
Inserted revision body by pull replication could be accessed from other codes. In case of the https://github.com/couchbase/couchbase-lite-android/issues/683 , `/_changes?include_docs=true` access revision body.

`if(rev.getBody() != null) rev.getBody().release();` was added to reduce memory usage during pull replication, but it should not be added here.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -515,8 +515,6 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
                     }
                 }
 
-                if(rev.getBody() != null) rev.getBody().release();
-
                 // Mark this revision's fake sequence as processed:
                 pendingSequences.removeSequence(fakeSequence);
             }


src/main/java/com/couchbase/lite/replicator/PullerInternal.java
```diff
@@ -515,8 +515,6 @@ public void insertDownloads(List<RevisionInternal> downloads) {
                     }
                 }
 
-                if(rev.getBody() != null) rev.getBody().release();
-
```
pasin: Will removing this line cause the memory regression issue during the pull replicating?

hideki: `Body.compact()` is called before reaching this line. So `Body` holds `json` value. I believe iOS implementation also holds `json` value at this point. However, as you mentioned, the peak memory usage might increase by deleting this line.

hideki: Additional Information: 
- Impact to memory usage by not calling `Body.release()` at here could be minimum.
- From functionality point of view, releasing json field makes functional deficit for Database Change Listener functionality.

PR #828, opened 2015-11-24
Fixed https://github.com/couchbase/couchbase-lite-android/issues/685

hideki:
Problem: `Database.compact()` method did not work as expected because of porting error.

Fixed porting error.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -1349,6 +1349,7 @@ public class Database implements StoreDelegate {
     public Map<String, Object> getAllDocs(QueryOptions options) throws CouchbaseLiteException {
 
         // TODO - This method is not fully implemented. Further work will be required.
+        //        https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLDatabase%2BInternal.m#L676-L739
 
         return store.getAllDocs(options);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1347,6 +1347,9 @@ public RevisionList unpushedRevisionsSince(String sequence,
 
     @InterfaceAudience.Private
     public Map<String, Object> getAllDocs(QueryOptions options) throws CouchbaseLiteException {
+
+        // TODO - This method is not fully implemented. Further work will be required.
```
pasin: Do we have an issue # for this? Could you please reference it here if we have one?

src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -2141,14 +2141,12 @@ protected boolean endTransaction(boolean commit) {
      * would require an expensive full tree traversal. Hopefully this way is good enough.
      */
     protected int pruneRevsToMaxDepth(int maxDepth) throws CouchbaseLiteException {
-        if (maxDepth == 0) {
+        if (maxDepth == 0)
             maxDepth = getMaxRevTreeDepth();
-        }
+
+        Log.e(TAG, "CBLDatabase: Pruning revisions to max depth %d...", maxDepth);
```
pasin: This shouldn't be an error log I guessed. Also, `CBLDatabase` is an iOS artifact, could you please change it to something else or remove it?

PR #845, opened 2015-12-02
Fixed #830 - Replication should stop immediately if error is not tran…

### LINKED ISSUE #830: Replication should stop immediately if error is not transient error.
depends on #573

hideki:
…sient error.

In case permanent error occurs during replication, replication keeps to replicate till it becomes IDLE state. Instead, it immediately stops replication if error is permanent.

Additional:
- Batcher.java: `Object.notify()` should be called in `synchronization` block
- Utils.java: code cleanup

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
@@ -92,7 +92,7 @@ abstract class ReplicationInternal implements BlockingQueueListener {
     protected static int PROCESSOR_DELAY = 500;
     protected static int INBOX_CAPACITY = 100;
     protected ScheduledExecutorService remoteRequestExecutor;
-    protected Throwable error;
+    private Throwable error; // use private to make sure if error is set through setError()
     private String remoteCheckpointDocID;
     protected Map<String, Object> remoteCheckpoint;
     protected AtomicInteger completedChangesCount;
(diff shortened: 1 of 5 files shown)


src/main/java/com/couchbase/lite/support/Batcher.java
```diff
@@ -146,7 +146,9 @@ public void flushAll() {
     public void clear() {
         unschedule();
         inbox.clear();
-        inbox.notify();
+        synchronized (inbox) {
```
pasin: Do we need to move `inbox.clear()` to inside synchronized block as multiple threads could access inbox?

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -450,27 +449,29 @@ public void onCompletion(HttpResponse httpResponse, Object result, Throwable e)
...
-        // state becomes IDLE.
-    }
+            // if permanent error, stop immediately
+            if(Utils.isPermanentError(this.error)) {
+                stop();
+            }
 
+            // iOS version sends notification from stop() method, but java version does not.
+            // following codes are always executed.
+            {
```
pasin: { here is not necessary.

src/main/java/com/couchbase/lite/support/Batcher.java
hideki: `inbox` is synchronized List. so it might not be necessary. 
`this.inbox = Collections.synchronizedList(new ArrayList<T>());`

hideki:
please review this again?

Issue #844, opened 2015-12-02
Port iOS mapping status code to HTTP status code for CouchbaseLiteException


pasin:
When setting a status code to CouchbaseLiteException, the status code needs to map to a HTTP status code.

https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLStatus.m#L75-L86

PR #846, opened 2015-12-02
Fixed #722 and #844 Support for upgrading from SQLite databases to ForestDB

### LINKED ISSUE #722: Migrating SQLite db to ForestDB
iOS Ticket: https://github.com/couchbase/couchbase-lite-ios/issues/870

pasin:
- When open the database, SQLite database can be upgrade to forestdb database (#722)

- Implement Status to HTTP Status conversion (#844)

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -1120,10 +1120,11 @@ public class Database implements StoreDelegate {
                     Status.INTERNAL_SERVER_ERROR);
 
         String storageType = options.getStorageType();
-        if (storageType == null)
+        if (storageType == null) {
             storageType = manager.getStorageType();
-        if (storageType == null)
-            storageType = DEFAULT_STORAGE;
+            if (storageType == null)
+                storageType = DEFAULT_STORAGE;
+        }
 
         Store primaryStore = createStoreInstance(storageType);
         if (primaryStore == null) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1225,8 +1226,17 @@ public synchronized void open(DatabaseOptions options) throws CouchbaseLiteExcep
         open = true;
 
         if (upgrade) {
-            // TODO: Perform database upgrading here:
-            // https://github.com/couchbase/couchbase-lite-java-core/issues/722
+            Log.i(TAG, "Upgrading to %s ...", storageType);
+            String dbPath = new File(path, "db.sqlite3").getAbsolutePath();
```
hideki: Is this upgrade method support only Android/Java v1.2.0 SQLite database? How about pervious version?

NitzDKoder:
@hideki @pasin Is this available in CBL 1.2?

pasin:
@NitzDKoder Yes, it's in CBL 1.2.

ddash1:
@pasin What is the mechanism to migrate existing data from SQLite to ForestDB, if any? Or we have to start fresh with ForestDB instance.

hideki:
Hi @ddash1 

Please open database with using `Manager. openDatabase(String name, DatabaseOptions options)`
https://github.com/couchbase/couchbase-lite-java-core/blob/master/src/main/java/com/couchbase/lite/Manager.java#L330

This blog post could help you.
http://blog.couchbase.com/2016/february/mobile-1.2-enable-forestdb

ddash1:
thanks @hideki, understood the new DB with ForestDB storage.
Not clear about how to migrate the existing data from SQLite to ForestDB?

pasin:
@ddash1 If you had an existing SQLite database file, when you call Manager.openDatabase to open that database with ForestDB storage specified, the data will be automatically migrate to the ForestDB storage.

ddash1:
@pasin, I ask because

http://blog.couchbase.com/2016/february/mobile-1.2-enable-forestdb
"Note that if you have a previous project using an earlier version of Couchbase Lite then **the existing database files are not converted**. This is where a SQLite-backed database will continue using SQLite by default because the format of the database type is established at the beginning of project creation. The existing database will always open with the same storage engine that created them. Stay tuned for future blogs that will address how to migrate that over."
https://github.com/couchbase/couchbase-lite-java-core/pull/846

pasin:
There are 2 ways that you can use ForestDB storage:

1. Set Manager.storageType to ForestDB and use Manager.getDatabase() method like the before (v1.1 and below) to create a new database or get an existing one. Setting storageType to the manager will set the default storage engine but that will only affect to new databases that you will create. It will not perform any migration on existing database.

2. Use the new openDatabase() and specify the storage type to ForestDB. This method will perform auto-migration from SQLite to ForestDB but not the other way around.

PR #853, opened 2015-12-08
Fixed #852 - Cookie is not set for /{db}/_revs_diff with one shot pus…

### LINKED ISSUE #852: Cookie is not set for /{db}/_revs_diff with one shot push replication
Steps to reproduce
1. Create user and generate Sync Gateway session ID by curl comand
2. Start One Shot Push replication after setting cookie with session ID which is obtained by step 1.
3. /{db}/_revs_diff fails with HTTP 401.

Screen Shot from HttpScoop
screen shot 2015-12-07 at 11 03 30 pm

Sync Gateway Log
```
2015-12-07T23:01:30.241-08:00 Cache: Received #1 ("_user/user1")
2015-12-07T23:01:30.241-08:00 Cache+:   Ignoring duplicate of #1
2015-12-07T23:03:03.373-08:00 HTTP:  #003: GET /db/_local/3cdc24237984c949d114ce7a797562b036e1a7dd  (as user1)
2015-12-07T23:03:03.373-08:00 HTTP: #003:     --> 404 missing  (0.3 ms)
2015-12-07T23:03:03.918-08:00 HTTP:  #004: POST /db/_revs_diff
2015-12-07T23:03:03.918-08:00 HTTP: #004:     --> 401 Login required  (0.2 ms)
```

NOTE: Cookie value is not set only first `/{db}/revs_diff` request. From second request, cookie is set in HTTP header.

hideki:
…h replication

- Problem: Sync Gateway returns response with `Set-Cookie:` for Sync Gateway Session ID. As HttpClient automatically updates Cookie with path for requested URL. Updated Cookie path does not match for next request.
- Solution: Update cookie value with existing cookie path.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/PersistentCookieStore.java
+++ b/src/main/java/com/couchbase/lite/support/PersistentCookieStore.java
@@ -23,6 +23,7 @@ import com.couchbase.lite.util.Log;
 
 import org.apache.http.client.CookieStore;
 import org.apache.http.cookie.Cookie;
+import org.apache.http.impl.cookie.BasicClientCookie2;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -39,6 +40,8 @@ import java.util.concurrent.ConcurrentHashMap;
 
 
 public class PersistentCookieStore implements CookieStore {
+
+    private static final String SYNC_GATEWAY_SESSION_COOKIE_NAME = "SyncGatewaySession";
 
     private static final String COOKIE_LOCAL_DOC_NAME = "PersistentCookieStore";
     private boolean omitNonPersistentCookies = false;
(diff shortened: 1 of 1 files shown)


pasin:
Taken from Hideki's comment in #852,

`Set-Cookie:` does not contain any path. HttpClient set same path with request URL. That causes mis-matching for next request with /db/_revs_diff .

Response Header Example 
```
curl -i --cookie "SyncGatewaySession=27c8af92b9dc54121b24a3b8f21c218c858beb11" -H "Content-Type: application/json" -X GET  http://localhost:4984/db/
HTTP/1.1 200 OK
Content-Length: 159
Content-Type: application/json
Server: Couchbase Sync Gateway/1.1.1
Set-Cookie: SyncGatewaySession=27c8af92b9dc54121b24a3b8f21c218c858beb11; Expires=Tue, 08 Dec 2015 23:16:53 UTC
Date: Tue, 08 Dec 2015 23:11:53 GMT

{"committed_update_seq":31,"compact_running":false,"db_name":"db","disk_format_version":0,"instance_start_time":1449616103704232,"purge_seq":0,"update_seq":31}
```

pasin:
Per https://www.ietf.org/rfc/rfc2109.txt spec, section 4.3.1, HttpClient does the right thing when there is no path setting at the Set-Cookie header. Does this mean the cookie returned for `/{db}/_revs_diff` should be from the cookie of `/db`?

pasin:
I understand it now. The current implementation is not completed. It stores cookie by name constructing from cookie name and domain. There is no path matching. I think we shouldn't change anything as we are moving to use the new HttpClient.

hideki:
@pasin Thank you for your research.

I believe HttpClient set cookies in header from list of cookies in CookieStore, and it seems it does not rely on Hash matching. I assume overwriting `path` value with URL path causes cookie mismatching.

pasin:
CookiePersistentStore doesn't implement cookie store per the spec. Cookies with different paths should be saved into the store independently. CookiePersistentStore just stores by domain & name which is not correct but might be working for Sync-Gateway with the workaround from this PR?.

I read HttpClient Cookie Store source code once, and if I remembered correctly, it does implement cookie path matching at a certain level. The cookie matching spec can be found here, https://tools.ietf.org/html/rfc6265.

PR #869, opened 2015-12-16
Fixed #695 Implemented Replicator.remoteUUID API

### LINKED ISSUE #695: Add replicator support for remote dbs with varying URLs
This is a parity feature with iOS.

Issue: https://github.com/couchbase/couchbase-lite-ios/issues/733
Commit: https://github.com/couchbase/couchbase-lite-ios/commit/2a75f83c1fd04410c9dde44a817f5e708492dd6c

pasin:
Implemented Replicator.remoteUUID used for representing the remote database instead of URL for the cases (e.g. P2P) that the URLs are not static.

#695

### DIFF
--- a/src/main/java/com/couchbase/lite/Manager.java
+++ b/src/main/java/com/couchbase/lite/Manager.java
@@ -678,6 +678,11 @@ public final class Manager {
                 }
             }
 
+            String remoteUUID = (String) properties.get("remoteUUID");
+            if (remoteUUID != null) {
+                repl.setRemoteUUID(remoteUUID);
+            }
+
             if (push) {
                 repl.setCreateTarget(createTarget);
             }
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -87,7 +87,8 @@ public class Replication implements ReplicationInternal.ChangeListener, NetworkR
         DOC_IDS,
         REQUEST_HEADERS,
         AUTHENTICATOR,
-        CREATE_TARGET
+        CREATE_TARGET,
+        REMOTE_UUID
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/main/java/com/couchbase/lite/Manager.java
```diff
@@ -678,6 +678,11 @@ public Replication getReplicator(Map<String, Object> properties) throws Couchbas
                 }
             }
 
+            String remoteUUID = (String) properties.get("remoteUUID");
+            if (remote != null) {
+                repl.setRemoteUUID(remoteUUID);
```
hideki: Needs to check if `remoteUUID` is `null`??

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -700,6 +701,31 @@ public void deleteCookie(String name) {
         replicationInternal.deleteCookie(name);
     }
 
+    /**
+     * Get the remote UUID representing the remote server.
+     */
+    @InterfaceAudience.Public
+    public String getRemoteUUID() {
+        return replicationInternal.getRemoteUUID();
```
hideki: `replicationInternal ` could be null?

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -700,6 +701,31 @@ public void deleteCookie(String name) {
...
+     * As a result, URL couldn't be used to represent the remote database when persistently
+     * identifying the replication. In such cases, set a remote unique identifier to represent
+     * the remote database.
+     *
+     * @param remoteUUID remote unique identifier
+     */
+    @InterfaceAudience.Public
+    public void setRemoteUUID(String remoteUUID) {
+        properties.put(ReplicationField.REMOTE_UUID, remoteUUID);
+        replicationInternal.setRemoteUUID(remoteUUID);
```
hideki: same here

src/main/java/com/couchbase/lite/Manager.java
pasin: I got wrong. It should check only remoteUUID actually. Will fix this.

pasin:
@hideki  as the fix per your review is very small, I rebased it to the original commit. I also rebased with the current master. Could you please re-review? Thanks!

Issue #682, opened 2015-07-20
Add "_conflicts" property to doc in map function when doc is in conflict


hideki:
https://github.com/couchbase/couchbase-lite-ios/issues/806

PR #876, opened 2015-12-18
Fixed #681 and #682 Catch up View.updateIndexes for SQLite View Storage

### LINKED ISSUE #681: Catch up latest (v1.1) View.updateIndex() implementation
Needs to implement iOS `viewsInGroup` and `updateIndexes: (NSArray*)views` to support prefix view indexing for both SQLite and ForestDB

pasin:
- Implemented group view index, views whose name has the same prefix before slash will bet indexed together #681.

- _conflict map function is implmeneted #682.

### DIFF
--- a/src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
+++ b/src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
@@ -68,6 +68,7 @@ public class SQLiteViewStore implements ViewStore, QueryRowStore {
     private int viewID;
     private View.TDViewCollation collation;
     private String _mapTableName;
+    private SQLiteViewStore curView; // Current view used when update index
 
     ///////////////////////////////////////////////////////////////////////////
     // Constructor
@@ -247,8 +248,6 @@ public class SQLiteViewStore implements ViewStore, QueryRowStore {
     @InterfaceAudience.Private
     public Status updateIndexes(List<ViewStore>inputViews) throws CouchbaseLiteException {
         Log.v(Log.TAG_VIEW, "Re-indexing view: %s", name);
-        assert (delegate.getMap() != null);
-
         if (getViewID() <= 0) {
             String msg = String.format("getViewID() < 0");
             throw new CouchbaseLiteException(msg, new Status(Status.NOT_FOUND));
@@ -269,20 +268,21 @@ public class SQLiteViewStore implements ViewStore, QueryRowStore {
             // Check whether we need to update at all,
             // and remove obsolete emitted results from the 'maps' table:
             long minLastSequence = dbMaxSequence;
-            long[] viewLastSequence = new long[inputViews.size()];
+            final long[] viewLastSequence = new long[inputViews.size()];
             int deletedCount = 0;
             int i = 0;
-            HashSet<String> docTypes = new HashSet<String>();
+            final HashSet<String> docTypes = new HashSet<String>();
             HashMap<String, String> viewDocTypes = null;
             boolean allDocTypes = false;
-            HashMap<Integer, Integer> viewTotalRows = new HashMap<Integer, Integer>();
-            ArrayList<SQLiteViewStore> views = new ArrayList<SQLiteViewStore>();
-            ArrayList<Mapper> mapBlocks = new ArrayList<Mapper>();
+            final HashMap<Integer, Integer> viewTotalRows = new HashMap<Integer, Integer>();
+            final ArrayList<SQLiteViewStore> views = new ArrayList<SQLiteViewStore>();
+            final ArrayList<Mapper> mapBlocks = new ArrayList<Mapper>();
 
             for (ViewStore v : inputViews) {
+                assert(v != null);
                 SQLiteViewStore view = (SQLiteViewStore)v;
                 ViewStoreDelegate delegate = view.getDelegate();
-                Mapper map = delegate.getMap();
+                Mapper map = delegate != null ? delegate.getMap() : null;
                 if (map == null) {
                     if (view == this) {
                         String msg = String.format("Cannot index view %s: " +
@@ -311,13 +311,13 @@ public class SQLiteViewStore implements ViewStore, QueryRowStore {
                 long last = view == this ? forViewLastSequence : view.getLastSequenceIndexed();
                 viewLastSequence[i++] = last;
                 if (last < 0) {
-                    String msg = String.format("last < 0 (%s)", last);
+                    String msg = String.format("last < 0 (%d)", last);
                     throw new CouchbaseLiteException(msg, new Status(Status.INTERNAL_SERVER_ERROR));
                 } else if (last < dbMaxSequence) {
                     if (last == 0)
                         view.createIndex();
                     minLastSequence = Math.min(minLastSequence, last);
-                    Log.v(Log.TAG_VIEW, "    %s last indexed at #%s", view.getName(), last);
+                    Log.v(Log.TAG_VIEW, "    %s last indexed at #%d", view.getName(), last);
 
                     String docType = delegate.getDocumentType();
                     if (docType != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -250,48 +255,115 @@ public void updateIndex() throws CouchbaseLiteException {
...
+            ArrayList<SQLiteViewStore> views = new ArrayList<SQLiteViewStore>();
+            ArrayList<Mapper> mapBlocks = new ArrayList<Mapper>();
+
+            for (ViewStore v : inputViews) {
+                SQLiteViewStore view = (SQLiteViewStore)v;
+                ViewStoreDelegate delegate = view.getDelegate();
+                Mapper map = delegate.getMap();
+                if (map == null) {
+                    if (view == this) {
+                        String msg = String.format("Cannot index view %s: " +
```
hideki: This does not make sense.
If view is null, `view.getDelegate()` should throw `NullPointerException`.
Also, theoretically, `delegate` could be `null`. `delegate.getMap()` might through `NullPointerExcepiton`.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -250,48 +255,115 @@ public void updateIndex() throws CouchbaseLiteException {
...
+                    String message = String.format("View '%s' not found in database",
+                            view.getName());
+                    Log.e(Log.TAG_VIEW, message);
+                    throw new CouchbaseLiteException(message, new Status(Status.NOT_FOUND));
+                }
+
+                int totalRows = view.getTotalRows();
+                viewTotalRows.put(viewID, totalRows);
+
+                long last = view == this ? forViewLastSequence : view.getLastSequenceIndexed();
```
hideki: `long last = view.getLastSequenceIndexed(); ` is simpler.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -250,48 +255,115 @@ public void updateIndex() throws CouchbaseLiteException {
...
-                result.setCode(Status.NOT_MODIFIED);
-                return;
+                success = true;
+                return new Status(Status.NOT_MODIFIED);
             }
 
-            // This is the emit() block, which gets called from within the
-            // user-defined map() block
+            Log.v(Log.TAG_VIEW, "Updating indexes of (%s) from #%s to #%s ...",
+                    viewNames(views), minLastSequence, dbMaxSequence);
```
hideki: `%s` for `long` value could cause unexpected error.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -308,9 +380,10 @@ public void emit(Object key, Object value) {
                         String[] args = {Long.toString(sequence), keyJson, valueJson};
                         store.getStorageEngine().execSQL(queryString(
                                 "INSERT INTO 'maps_#' (sequence, key, value) VALUES(?,?,?)"), args);
+                        insertedCount.incrementAndGet();
                     } catch (Exception e) {
                         Log.e(Log.TAG_VIEW, "Error emitting", e);
-                        // find a better way to propagate this back
+                        throw new RuntimeException(e);
```
hideki: Which `try catch block` does catch this `RuntimeException` ?

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -358,45 +434,68 @@ public void emit(Object key, Object value) {
...
-                    String[] selectArgs2 = {Long.toString(docID), Long.toString(minLastSequence)};
-
                     Cursor cursor2 = null;
                     try {
+                        String[] selectArgs2 = {Long.toString(docID), Long.toString(minLastSequence)};
                         cursor2 = store.getStorageEngine().rawQuery(
                                 "SELECT revid, sequence FROM revs "
                                         + "WHERE doc_id=? AND sequence<=? AND current!=0 AND deleted=0 "
-                                        + "ORDER BY revID DESC "
-                                        + "LIMIT 1", selectArgs2);
```
hideki: Question: Could multiple revisions be indexed? (Regarding to delete `LIMIT 1`)

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
```diff
@@ -406,61 +505,79 @@ public void emit(Object key, Object value) {
...
+                            String viewDocType = viewDocTypes.get(view.getName());
+                            if (viewDocType != null && !viewDocType.equals(docType))
+                                continue; // skip; view's documentType doesn't match this doc
+                        }
+                    }
+                    Log.v(Log.TAG_VIEW, "#%s: map '%s' for view %s...",
+                            sequence, docID, view.getName());
+                    try {
+                        emitBlock.setSequence(sequence);
+                        delegate.getMap().map(curDoc, emitBlock);
```
hideki: I think `map` function is per View. Retrieving `map` function from `delegate` might apply different `map` function??

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
hideki: Could be cached by Line 550.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: Practically the NullPointerException will not occurred in both cases. 
1. There will be at least one view.
2. The delegate is set to `this` in the `View`'s constructor. 
However it is a good practice to guard that. I will make a fix for this.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: It's simpler, but it's better to not calling `view.getLastSequenceIndexed()` to query value from the database again. So let's keep it that way.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: I think java does the conversion. The original code in View.updateIndex() also does that. Please see line #355 above. Anyway I will change all of them.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: That is my attention. Do we see any issues?

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: This SQL statement is for getting all conflicting revisions to construct `_conflicts` for line 532.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: Big thanks! for catching this. I was supposed to get the map from the `mapBlocks` (line 280). I copied this piece from the original code and forgot to make changes.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
pasin: After thought about this, I prefer to not check if view is null as 
1. To let it failed as it means that we have a more serious bug somewhere else.
2. The case shouldn't be happening.

hideki: `view == null` check is not necessary here.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
hideki: sorry. I thought `getLastSequenceIndexed ()` method is a simple getter method.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
hideki: http://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html

```
's', 'S'	general	If the argument arg is null, then the result is "null". If arg implements Formattable, then arg.formatTo is invoked. Otherwise, the result is obtained by invoking arg.toString().
```

`long` -> `Long` -> `Long.toString()` could be happened here.

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
hideki: I just worried RuntimeException could not be caught by anything.
``
RuntimeException and its subclasses are unchecked exceptions. Unchecked exceptions do not need to be declared in a method or constructor's throws clause if they can be thrown by the execution of the method or constructor and propagate outside the method or constructor boundary
```

src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
hideki: Thanks!

pasin:
@hideki I fixed per your code review and couple issues in emit block. Please re-review. Thanks!

Issue #870, opened 2015-12-16
Sync the code with iOS CBL_ReplicatorSettings


pasin:
iOS refactors and extracts replicators around replicator settings into CBL_ReplicatorSettings. It might be a good idea to do the same thing.

PR #892, opened 2015-12-23
Fixed #535 : setting documentIDs not working in Push replication

### LINKED ISSUE #535: The replicator's setDocIds method doesn't filter the documents to push
The Replication class has a setDocIds method to provide a list of doc ids to only sync those docs in a push replication.

I believe this method is not documented on the mobile portal yet. The only time I saw it was used is in the PhotoDrop app @pasin wrote to selectively push the selected pictures.

I was wondering if it worked with Lite -> SG replication as well.

See this example to reproduce https://github.com/jamiltz/ReplicationFilter#array-of-document-ids

pasin:
Implemented documentID filtering for push replication.

#535

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -32,7 +32,6 @@ import java.util.concurrent.TimeUnit;
  * The external facade for the Replication API
  */
 public class Replication implements ReplicationInternal.ChangeListener, NetworkReachabilityListener {
-
     /**
      * Enum to specify which direction this replication is going (eg, push vs pull)
      *
@@ -172,7 +171,6 @@ public class Replication implements ReplicationInternal.ChangeListener, NetworkR
      */
     @InterfaceAudience.Public
     public void start() {
-
         if (replicationInternal == null) {
             initReplicationInternal();
         } else {
@@ -199,7 +197,6 @@ public class Replication implements ReplicationInternal.ChangeListener, NetworkR
                 }
             });
         }
-
 
         replicationInternal.triggerStart();
     }
@@ -380,7 +377,6 @@ public class Replication implements ReplicationInternal.ChangeListener, NetworkR
      */
     @Override
     public void changed(ChangeEvent event) {
-
         // forget cached IDs (Should be executed in workExecutor)
         if (pendingDocIDs != null) {
             db.runAsync(new AsyncTask() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -641,16 +634,16 @@ public void setFilter(String filterName) {
      * Sets the documents to specify as part of the replication.
      */
     @InterfaceAudience.Public
-    public void setDocIds(List<String> docIds) {
+    public void setDocumentIDs(List<String> docIds) {
```
hideki: setDocIds is public API, and it is documented with 
```
public List<String> getDocIds()
public void setDocIds(List<String> docIds)
```
http://developer.couchbase.com/documentation/mobile/1.1.0/develop/references/couchbase-lite/couchbase-lite/replication/replication/index.html#string-docids--get-set-

This method signature should not be changed.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -965,6 +967,29 @@ public void setFilter(String filterName) {
         this.filterName = filterName;
     }
 
+    /**
+     * Get replicator filter. If the replicator is a pull replicator,
+     * calling this method will return null.
+     *
+     * This is equivalent to CBL_ReplicatorSettings's compilePushFilterForDatabase:status:.
+     */
+    public ReplicationFilter compilePushReplicationFilter() {
```
hideki: should this method be declared in PusherInternal?

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -974,14 +999,14 @@ public void setFilter(String filterName) {
     /**
      * Gets the documents to specify as part of the replication.
      */
-    public List<String> getDocIds() {
+    public List<String> getDocumentIDs() {
```
hideki: If we keep current API name, this method signature should not be changed.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -974,14 +999,14 @@ public void setFilter(String filterName) {
...
         return documentIDs;
     }
 
     /**
      * Sets the documents to specify as part of the replication.
      */
-    public void setDocIds(List<String> docIds) {
+    public void setDocumentIDs(List<String> docIds) {
```
hideki: same with above

src/main/java/com/couchbase/lite/replicator/Replication.java
pasin: iOS has documentIDs. I will change back to DocIds and let @zgramana review the discrepency.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
pasin: No because it is used in Replication.java as well. We should adopt ReplicationSettings from iOS.

https://github.com/couchbase/couchbase-lite-java-core/issues/870

pasin:
I fixed the API. Could you please re-review?

Issue #899, opened 2016-01-06
Batcher not always execute tasks with work executor


pasin:
Batcher only executed delayed tasks in work executor. In queueObjects() and flushAll(), tasks get executed outside work executor. This could result to unexpected behavior of the replicator especially the pull replicator which has change tracker operates on its own thread and adds some tasks to the batcher.

pasin:
Related issue: https://github.com/couchbase/couchbase-lite-android/issues/697

PR #915, opened 2016-01-11
Fix batcher synchronization

pasin:
- Changed to synchronize over mutex.
- When the scheduled task is ready to be processed but there is still a thread keep adding objects to the batcher, give the work executor more chance to work on the task by calling mutex.wait() with a small dealy in queueObjects.
- Fixed incompatible runloop capability when scheduing the task after long pause - instead of schedule with 0 delay, schedule with some small resonable delay to allow more objects to be added to the batcher if available.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/Batcher.java
+++ b/src/main/java/com/couchbase/lite/support/Batcher.java
@@ -28,12 +28,16 @@ public class Batcher<T> {
     private int capacity = 0;
     private long delay = 0;
     private List<T> inbox;
+
     private boolean scheduled = false;
     private long scheduledDelay = 0;
+    private ScheduledFuture pendingFuture = null;
+
     private BatchProcessor<T> processor;
     private long lastProcessedTime = 0;
-    private ScheduledFuture pendingFuture = null;
-    private Object mutex = new Object();
+
+    private final Object mutex = new Object();
+    private final Object processMutex = new Object();
 
     ///////////////////////////////////////////////////////////////////////////
     // Constructors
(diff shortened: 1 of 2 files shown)


pasin:
I just rebased with the current master and added a small enhancement. Ready for the review now.

src/main/java/com/couchbase/lite/support/Batcher.java
```diff
@@ -55,7 +65,7 @@ public Batcher(ScheduledExecutorService workExecutor,
         this.processor = processor;
         this.inbox = new ArrayList<T>();
         this.scheduled = false;
-        this.lastProcessedTime = System.currentTimeMillis();
+        this.lastProcessedTime = 0;
```
hideki: Is this change that makes batcher immediately execute after create new instance of Batcher and then added tasks smaller than capacity size?

src/main/java/com/couchbase/lite/support/Batcher.java
```diff
@@ -55,7 +65,7 @@ public Batcher(ScheduledExecutorService workExecutor,
         this.processor = processor;
         this.inbox = new ArrayList<T>();
```
hideki: From line 66 - 68, variables are initialized with their declaration. We could remove line 66 - 68. And inbox could be initialized with its declaration.

src/main/java/com/couchbase/lite/support/Batcher.java
pasin: As discussed, setting lastProcessedTime to the current time here will result to unpredictable result depending on time to create the batcher and time to queue the object. For more info, iOS also sets the value to zero and let the schedule objects to decide what the delay should be.

src/main/java/com/couchbase/lite/support/Batcher.java
pasin: OK.

pasin:
With the recent commit, all unit tests were passed on Jenkin server. Could you please review?

pasin:
Reference: https://github.com/couchbase/couchbase-lite-java-core/issues/899

PR #929, opened 2016-01-19
Fix class cast exception when revpos is a Double

reisub:
It seems that _revpos_ can sometimes be a Double and this causes an exception.
If we change the cast to a Number, it all works out well because `Number` also has the `intValue()` method.

We've been using a build with this fix in production in the last couple of months without any issues so I'm pretty sure it's safe. The only thing I'm unsure is why this exception happens in the first place.


hideki:
@reisub Thank you very much for your PR.

PR #942, opened 2016-01-25
Fix null id in bulk docs response

jamiltz:
To reproduce the issue:
```
curl -X POST http://172.20.6.40:5984/cblite-test/_bulk_docs -d '{"docs": [{"city": "brighton"}]}'
```
Response:
```
[{"rev":"1-f167f0e9f44ce7cf3c92d53910fb9179","id":null,"ok":true}]
```

Fix: return the id of the revision created. The id will be the one provided by the user or the one generated in the write operation.

### DIFF
--- a/src/main/java/com/couchbase/lite/router/Router.java
+++ b/src/main/java/com/couchbase/lite/router/Router.java
@@ -1199,7 +1199,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
                         if (status.isSuccessful()) {
                             result = new HashMap<String, Object>();
                             result.put("ok", true);
-                            result.put("id", docID);
+                            result.put("id", rev.getDocID());
                             if (rev != null) {
                                 result.put("rev", rev.getRevID());
                             }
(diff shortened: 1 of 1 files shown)


hideki:
Hi @jamiltz,
Thank you very much for finding the problem and sending your PR! 
Hideki

jamiltz:
I managed to get the couchbase-lite-android-liteserv to compile after lots of trial and error :) I'll open a PR on the README of that repo to share the steps I followed. Or do you think it should go somewhere else? I have another question, I see that the master branch is continuously deployed as [maven modules](http://files.couchbase.com/maven2/com/couchbase/lite/) which is really great! I'm wondering how I could get hold of the JAR files. Can I build them from source with couchbase-lite-android-liteserv or another repo or, ideally, is there a way to pull only the JARs from the Couchbase maven repo into a local directory?

hideki:
> I'll open a PR on the README of that repo to share the steps I followed. 

README for couchbase-lite-android is best place or Wiki. Thanks!

> I see that the master branch is continuously deployed as maven modules which is really great!

Recently, we fixed Jenkins server. This is great to me too. Till end of 1.2 release, I recommend to use 1.2.0-xxx as they have more stable than master. 

> Can I build them from source with couchbase-lite-android-liteserv or another repo or, ideally, is there a way to pull only the JARs from the Couchbase maven repo into a local directory?

You can download zipped jar files from following links.
http://latestbuilds.hq.couchbase.com/couchbase-lite-android/release/
http://latestbuilds.hq.couchbase.com/couchbase-lite-java/release/

jamiltz:
Thanks Hideki! Those links are super useful. Ok found the wiki.

PR #956, opened 2016-01-27
Fixed #952 - view rev query order by deleted asc

### LINKED ISSUE #952: Views broken with concurrent update and delete
Hi guys....

We've spent about a week and a half with a very annoying issue with this library.  We moved our server from CouchDB to Sync Server to resolve it (finding other issues like https://github.com/couchbase/sync_gateway/issues/1505 in the meantime).  And changed out server logic a few times, and watched the change feeds. But the bug still persists and is quite annoying.  Now we have clearly tracked it down to the java couchbase lite library.

This is what we do:

- We have an document shared on multiple devices and our server.  Base revision 1-x.
- We have a view function on the client to show all docs of type X, it displays this document.
- The two devices make changes and sync them concurrently (either very good timing, or taking one device offline, editing, then bringing it back online).
-  We now have a conflict between versions 2-a and 2-b on the server.
- So far the view looks good and shows the same (arbitrary) revision on both clients. (View updated well)
- Now, we run a conflict resolution script that marks 2-a as deleted (3-c) and updates 2-b with the merged data (3-d)
- Both these changes are synced in the same _changes query to the couchbase lite client and processed in one batch
- If rev of 3-d > rev of 3-c, everything works fine and we see the merged object in the view
- If rev of 3-d < rev of 3-c, the object is considered deleted on the client, and disappears from the view
  - THIS is the problem

More details:

- All data looks proper on the server
- Also, if we restart the app, the document is still missing from the view (I guess this is not updated again?)
- The custom view function for the view is never triggered by the last update (for either document rev 3-c or 3-d).  We have some log messages there and nothing happens in this case.
- If we wipe the local DB clean, and resync from the server, the document will appear correctly (I guess we trigger different client logic?)
- If we pull out the sqllite store from the client (even with a broken state) and look at the revision table, we see both rev 3-c and 3-d marked as current, and 3-c as deleted, which is proper
- If we look at the maps_N table for this view, it is empty, the document is not there
- Looking at winningRevIDOfDocNumericID (SQLiteStore.java:1971), the query there will calculate the correct winning revision from the data store.

Currently we trigger this only on the server, but I doubt it would be much different if the conflict resolution occurred on a client.  The main point is that one branch is deleted and the second updated in the same batch. And it leave the client in a broken state.

Given that the main use case for Couchbase is the conflict resolution (offline sync), and I figure we are doing a typical form of conflict resolution (update one branch, delete the other), I am quite surprised that this issue hasn't popped up earlier. And seems quite critical to our use case.

Currently, two developers have spent an entire week on the client and server hunting this issue. And this unreliability is threatening to block the adoption of offline sync by our company, which is annoying, as in theory this software matches our use case perfectly.

Please let me know if you need any more help to reproduce this bug.

alexlehn:
Small change to SQLiteViewStore updateIndexes to ensure the queried revs are provided with the non deleted first for #952.

### DIFF
--- a/src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
+++ b/src/main/java/com/couchbase/lite/store/SQLiteViewStore.java
@@ -401,7 +401,10 @@ public class SQLiteViewStore implements ViewStore, QueryRowStore {
                         docTypes.toArray(new String[docTypes.size()]));
                 sql.append("AND doc_type IN (").append(docTypesString).append(") ");
             }
-            sql.append("AND revs.doc_id = docs.doc_id ORDER BY revs.doc_id, revid DESC");
+            // order result by deleted ASC so if multiple revs returned the non deleted are the first
+            // NOTE: Views broken with concurrent update and delete
+            // https://github.com/couchbase/couchbase-lite-java-core/issues/952
+            sql.append("AND revs.doc_id = docs.doc_id ORDER BY revs.doc_id, deleted ASC, revid DESC");
             String[] selectArgs = {Long.toString(minLastSequence)};
             cursor = store.getStorageEngine().rawQuery(sql.toString(), selectArgs);
 


hideki:
Hi @alexlehn,
Thank you very much for your PR for both java-core and unit-test.

Did you have chance to sign the contributor license agreement? If not, please sign it to merge your PR?
https://github.com/couchbase/couchbase-lite-ios/wiki/Sign-the-contributor-license-agreement

Note: You might encounter account issue to sign:
https://github.com/couchbase/couchbase-lite-java-core/issues/952#issuecomment-175694403

Thanks!

alexlehn:
Hi @hideki

Yes I've signed the CLA username is same as my github name.

hideki:
FYI: I'd like to merge this PR after we merge release/1.2.0 branch into master. Please wait for few more days. Thanks!

hideki:
Note to Self: iOS is already addressed this issue. It seems this is porting error.
https://github.com/couchbase/couchbase-lite-ios/blob/5ab375c0587eb0e09407d24ddd9a9d045f9345e6/Source/CBL_SQLiteViewStorage.m#L387-L388

hideki:
@alexlehn thank you very much for your contribution. And sorry for taking time to merge your PR.

Thank you!!!

PR #966, opened 2016-01-30
Fixed #963 - Cannot pull changes after sync_gateway goes offline, the…

### LINKED ISSUE #963: Cannot pull changes after sync_gateway goes offline, then online
I am testing TodoLite Android

1. I create a list and share it between 2 devices
2. I add a few task as see that they are shared correctly
3. I take the db offline via :4985/db/_offline
```
curl -X POST http://localhost:4985/todos/_offline
```
4. I add a few tasks on one client
5. I take the db online via :4985/db/_online
```
curl -X POST http://localhost:4985/todos/_online
```

Expected: I would expect the offline tasks to get pushed to sync_gateway and pulled to the other device.

Actual: Offline tasks do not replicate to the other device.

hideki:
…n online

Root Cause : see https://github.com/couchbase/couchbase-lite-java-core/issues/963#issuecomment-177007753

Solution: In case http status is error and  mode is one shot, should not continue to process response for `_changes` API

TEST: I used GrocerySync and SG 1.2 for test.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
+++ b/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
@@ -326,7 +326,8 @@ public class ChangeTracker implements Runnable {
                 Log.v(Log.TAG_CHANGE_TRACKER, "%s: Making request to %s", this, maskedRemoteWithoutCredentials);
                 HttpResponse response = httpClient.execute(request);
                 StatusLine status = response.getStatusLine();
-                if (status.getStatusCode() >= 300 && !Utils.isTransientError(status)) {
+                if ((status.getStatusCode() >= 300 && !Utils.isTransientError(status)) ||
+                        status.getStatusCode() >= 300 && mode != ChangeTrackerMode.LongPoll) {
                     Log.e(Log.TAG_CHANGE_TRACKER, "%s: Change tracker got error %d", this, status.getStatusCode());
                     this.error = new HttpResponseException(status.getStatusCode(), status.getReasonPhrase());
                     break;
@@ -425,7 +426,8 @@ public class ChangeTracker implements Runnable {
                         if (entity != null) {
                             try {
                                 entity.consumeContent();
-                            } catch (IOException e) { }
+                            } catch (IOException e) {
+                            }
                         }
                     }
                 }
--- a/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
+++ b/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
@@ -326,13 +326,17 @@ public class ChangeTracker implements Runnable {
                 Log.v(Log.TAG_CHANGE_TRACKER, "%s: Making request to %s", this, maskedRemoteWithoutCredentials);
                 HttpResponse response = httpClient.execute(request);
                 StatusLine status = response.getStatusLine();
-                if ((status.getStatusCode() >= 300 && !Utils.isTransientError(status)) ||
-                        status.getStatusCode() >= 300 && mode != ChangeTrackerMode.LongPoll) {
+                // In case response status is Error, ChangeTracker stops here
+                // except mode is LongPoll and error is transient.
+                if (status.getStatusCode() >= 300 &&
+                        ((mode == ChangeTrackerMode.LongPoll && !Utils.isTransientError(status)) ||
+                                mode != ChangeTrackerMode.LongPoll)) {
                     Log.e(Log.TAG_CHANGE_TRACKER, "%s: Change tracker got error %d", this, status.getStatusCode());
                     this.error = new HttpResponseException(status.getStatusCode(), status.getReasonPhrase());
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
```diff
@@ -326,7 +326,8 @@ public void process(HttpRequest request, HttpContext context) throws HttpExcepti
                 Log.v(Log.TAG_CHANGE_TRACKER, "%s: Making request to %s", this, maskedRemoteWithoutCredentials);
                 HttpResponse response = httpClient.execute(request);
                 StatusLine status = response.getStatusLine();
-                if (status.getStatusCode() >= 300 && !Utils.isTransientError(status)) {
+                if ((status.getStatusCode() >= 300 && !Utils.isTransientError(status)) ||
+                        status.getStatusCode() >= 300 && mode != ChangeTrackerMode.LongPoll) {
```
pasin: 2 things:
1. I checked both iOS and .NET and this change will make the code divert from both platform.
2. I understand that this will make continuous pull replicator stops right away without having change tracker retrying.

pasin:
> One of problems: In case of error, response from SG is not JSON Array. But ChangeTracker expects JSON Array, it causes infinite loop.

@hideki I couldn't find how above issue gets handled in this PR. Can you provide more information?

hideki:
> Can you provide more information?

This issue is not caused by retry logic. Retry logic is handle by Pull replicator. This PR does not change any retry logic. The issue is caused by entering infinite-loop at [ChangeTracker.java Line 389](https://github.com/couchbase/couchbase-lite-java-core/blob/release/1.2.0/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java#L389) because response body is JSON without  Array. See blow message body.
```json
{"error":"Service Unavailable","reason":"DB is currently under maintenance"} 
```
Initial value of ChangeTracker's mode variable is  [OneShot] (https://github.com/couchbase/couchbase-lite-java-core/blob/d704e12d973a1dcd763dd897ad34f0fee3950b5f/src/main/java/com/couchbase/lite/replicator/PullerInternal.java#L135-L139).
Some reason (I need to investigate this more), ChangeTracker is running with OneShot mode without switching to LongPoll mode for retry. Then, ChangeTracker enters infinite loop that is described before.

This PR makes ChangeTracker immediately stops if HTTP Status is Error in case mode is OneShot.

src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
hideki: 1. It is hard to compare with iOS ChangeTracker.m. .NET v1.2 is same with current version (without this change). https://github.com/couchbase/couchbase-lite-net/blob/release/1.2/src/Sharpened/Replicator/ChangeTracker.cs#L358
2. PullReplicator does not stop. ChangeTracker stops immediately if ChangeTracker is running with OneShot mode. (I need to investigate more relationship between PullReplicator continuous/oneshot mode and ChangeTracker's LongPoll/OneShot mode).

hideki:
@pasin, Please ping me if you need further information.

src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
pasin: For 2, I actually mean the change tracker (before upgrading to longpoll) of the continuous pull replicator will not do the retry. I may be ok.

pasin:
@hideki I understand what you describe but

1. Either normal or longpoll mode, change tracker will return the same format. In case of 503 error, ChangeTracker in longpoll mode also gets the same error massage in the body but it can handle that. Why can't we process the body the same way between normal and longpoll mode?

2. When we know that we have an error for both one-shot and longpoll mode, can we skip parsing the body (I think this is what iOS is current doing)?

3. If we really need the while loop to find an array, is it possible to check the end of content and exit the loop?

hideki:
1. I have same question. I will investigate it.
2. As you wrote, longpoll mode also should skip parsing.
3. I will apply your suggestion.

src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
hideki: @pasin, sorry, you are right. this check was not correct. Thank you very much for catch my coding mistake.

hideki:
Hi @pasin,

1. Both OneShot and LongPoll modes could use same parsing logic. But I don't have enough confidence to change code just before 1.2.0 release. So I will update parsing logic for v1.3 release. https://github.com/couchbase/couchbase-lite-java-core/issues/969
2. As you suggested, I found problem in the error condition check in initial PR. I fixed it.
3. In case mode is LongPoll and HttpStatus is error, skip to parse response body.
4. I also added end of token check for OneShot mode parsing code.

Please review latest PR?

Thanks!

PR #968, opened 2016-01-30
Fixed #964 - ToDoLite crashes during clear Cookie

### LINKED ISSUE #964: ToDoLite crashes during clear Cookie
It seems it requires to catch NOT_FOUND exception, and ignore it for this scenario.
See stack trace:
https://gist.github.com/sethrosetter/6fcb15e79e237a53e056

hideki:
- PersistentCookieStore implements CookieStore, and overwrite `clear()` method which is no Exception throws declared. Current code uses RuntimeException to throw Exception even though Exception throws is not declared. However it is bad idea. No caller methods catches `RuntimeException`, then causes application crash. Exception was thrown because cookie was not stored before. So It simply makes clear() method not throw `RuntimeException`.

- During debuting this issue, I also encountered that `Document.getCurrentRevision()` returns `null`. I added null check for `getCurrentRevision()`. Switching user in ToDoLite sample makes switch database, that might be root cause of this issue. Need to investigate ToDoLite also.

### DIFF
--- a/src/main/java/com/couchbase/lite/Document.java
+++ b/src/main/java/com/couchbase/lite/Document.java
@@ -238,7 +238,8 @@ public class Document {
      */
     @InterfaceAudience.Public
     public Object getProperty(String key) {
-        if (getCurrentRevision().getProperties().containsKey(key)) {
+        if (getCurrentRevision() != null &&
+                getCurrentRevision().getProperties().containsKey(key)) {
             return getCurrentRevision().getProperties().get(key);
         }
         return null;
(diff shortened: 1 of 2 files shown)


pasin:
I totally understand the fix but I wonder what is causing not_found exception while putting null to a local doc id?

hideki:
`deleteLocalDocument()` throws `CouchbaseLiteException` With `NOT_FOUND` if document does not exist for specified document id in local document.
- https://github.com/couchbaselabs/couchbase-lite-java-forestdb/blob/master/src/main/java/com/couchbase/lite/store/ForestDBStore.java#L1278
- https://github.com/couchbase/couchbase-lite-java-core/blob/4586f6a8f8b3ea8cbdb1362196eb5cb00d9b2969/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1935

PR #1017, opened 2016-02-10
Fixed #1016 - Static Analyze: Resource management issues

### LINKED ISSUE #1016: Static Analyze: Resource management issues
- I/O resource opened but not safely closed

hideki:
- I/O resource opened but not safely closed

### DIFF
--- a/src/main/java/com/couchbase/lite/BlobStore.java
+++ b/src/main/java/com/couchbase/lite/BlobStore.java
@@ -337,34 +337,6 @@ public class BlobStore {
         return result;
     }
 
-    public static BlobKey keyForBlobFromFile(File file) {
-        MessageDigest md;
-        try {
-            md = MessageDigest.getInstance("SHA-1");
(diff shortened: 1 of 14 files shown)


src/main/java/com/couchbase/lite/BlobStore.java
```diff
@@ -534,33 +469,30 @@ public boolean storeBlob(byte[] data, BlobKey outKey) {
...
+                fos.close();
+            } catch (IOException e) {
             }
         }
         return true;
     }
 
     private static byte[] getBytesFromFile(File file) throws IOException {
-        InputStream is = null;
+        InputStream is = new FileInputStream(file);
```
zgramana: The general best practice for native resource allocation is to minimize its lifespan: initialize just before you need it, and release it just after you finish with it. Why not move this line down to just before line 577/509, and reduce the scope of the `try`/`finally` block to just the `while` loop?

hideki:
I need to rework for this. Close this PR.

PR #1034, opened 2016-02-14
Fixed #1033 - Requests pending with POST _bulk_docs

### LINKED ISSUE #1033: Requests pending with POST _bulk_docs
It seems that simultaneous _bulk_docs requests (with 200 docs) on different databases (db1/db2) are causing the Listener to hang, possibly a side effect of #1030. When inserting docs individually, the behaviour is not observed.

Logs: https://gist.github.com/jamiltz/8959562dc848b0bae209

Thread dump: https://gist.github.com/jamiltz/76daf479c682d48cfdce

hideki:
##### Problem:
- _bulk_docs: Calling synchronized(db) in transaction wait for db lock with holding SQLite connection
- other request: waiting for SQLConnection with holding db lock.

##### Solution:
Router.java:
- Eliminate all synchronized block from Router as most of database access are protected by Transaction.
- Use runInTransaction for putLocalRevision() which is not inTransaction

Database.java:
- make compact() thread-safe

Concern:
- Database.delete() and Database.close()'s thread-safety. Need to review later.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -251,7 +251,9 @@ public class Database implements StoreDelegate {
      */
     @InterfaceAudience.Public
     public void compact() throws CouchbaseLiteException {
-        store.compact();
+        synchronized (store) {
+            store.compact();
+        }
         garbageCollectAttachments();
     }
 
--- a/src/main/java/com/couchbase/lite/router/Router.java
+++ b/src/main/java/com/couchbase/lite/router/Router.java
@@ -1014,9 +1014,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
         if (getQuery("rev") != null) {
             return new Status(Status.BAD_REQUEST);  // CouchDB checks for this; probably meant to be a document deletion
         }
-        synchronized (db) {
-            db.delete();
-        }
+        db.delete();
         return new Status(Status.OK);
     }
 
@@ -1176,6 +1174,8 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
 
         final Status status = new Status(Status.OK);
         final List<Map<String, Object>> results = new ArrayList<Map<String, Object>>();
+        // Transaction provide synchronized feature by SQLiteDatabase.
+        // In the transaction block, should not use `synchronized` block
         boolean ret = db.getStore().runInTransaction(new TransactionalTask() {
             @Override
             public boolean run() {
@@ -1190,7 +1190,6 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
                         if (noNewEdits) {
                             rev = new RevisionInternal(docBody);
                             if (rev.getRevID() == null || rev.getDocID() == null || !rev.getDocID().equals(docID)) {
-                                //status = new Status(Status.BAD_REQUEST);
                                 status.setCode(Status.BAD_REQUEST);
                             } else {
                                 List<String> history = Database.parseCouchDBRevisionHistory(doc);
@@ -1210,7 +1209,6 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
                                 result.put("rev", rev.getRevID());
                             }
                         } else if (allOrNothing) {
-                            //return status;  // all_or_nothing backs out if there's any error
                             return false;
                         } else if (status.getCode() == Status.FORBIDDEN) {
                             result = new HashMap<String, Object>();
@@ -1264,6 +1262,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
 
         // Look them up, removing the existing ones from revs:
         try {
+            // query db only, not necessary to be syncrhonized
             db.findMissingRevisions(revs);
         } catch (SQLException e) {
             Log.e(Log.TAG_ROUTER, "Exception", e);
@@ -1301,6 +1300,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
     public Status do_POST_Document_compact(Database _db, String _docID, String _attachmentName) {
         Status status = new Status(Status.OK);
         try {
+            // Make Database.compact() thread-safe
             _db.compact();
         } catch (CouchbaseLiteException e) {
             status = e.getCBLStatus();
@@ -1341,6 +1341,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
         Future future = db.runAsync(new AsyncTask() {
             @Override
             public void run(Database database) {
+                // purgeRevisions uses transaction internally.
                 Map<String, Object> purgedRevisions = db.purgeRevisions(docsToRevs);
                 asyncApiCallResponse.add(purgedRevisions);
             }
@@ -1563,6 +1564,7 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
             Log.v(Log.TAG_ROUTER, "Filter params=" + changesFilterParams);
         }
 
+        // changesSince() is query only. not required synchronized
         RevisionList changes = db.changesSince(since, options, changesFilter, changesFilterParams);
 
         if (changes == null) {
@@ -1623,27 +1625,14 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
                 String revID = getQuery("rev");  // often null
                 RevisionInternal rev = null;
                 if (isLocalDoc) {
+                    // query only -> not required synchronized
                     rev = db.getLocalDocument(docID, revID);
                 } else {
-                    /*
-                    rev = db.getDocument(docID, revID, options);
-                    // Handle ?atts_since query by stubbing out older attachments:
-                    //?atts_since parameter - value is a (URL-encoded) JSON array of one or more revision IDs.
-                    // The response will include the content of only those attachments that changed since the given revision(s).
-                    //(You can ask for this either in the default JSON or as multipart/related, as previously described.)
-                    List<String> attsSince = (List<String>) getJSONQuery("atts_since");
-                    if (attsSince != null) {
-                        String ancestorId = db.findCommonAncestorOf(rev, attsSince);
-                        if (ancestorId != null) {
-                            int generation = RevisionInternal.generationFromRevID(ancestorId);
-                            db.stubOutAttachmentsIn(rev, generation + 1);
-                        }
-                    }
-                    */
                     boolean includeAttachments = options.contains(TDContentOptions.TDIncludeAttachments);
                     if (includeAttachments) {
                         options.remove(TDContentOptions.TDIncludeAttachments);
                     }
+                    // query only -> not required synchronized
                     rev = db.getDocument(docID, revID, true);
                     if (rev != null) {
                         rev = applyOptionsToRevision(options, rev);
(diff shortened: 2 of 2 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -251,7 +251,9 @@ public void addChangeListener(ChangeListener listener) {
      */
     @InterfaceAudience.Public
     public void compact() throws CouchbaseLiteException {
-        store.compact();
+        synchronized (store) {
+            store.compact();
+        }
```
pasin: There is nothing wrong with this synchronized. I checked the Database.java code and found that some code path does synchronized (store) but some not. If the intension to make Database thread safe, all of those `synchronized (store)` and `not synchronized (store)` should be revised.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1264,6 +1262,7 @@ public Status do_POST_Document_revs_diff(Database _db, String _docID, String _at
 
         // Look them up, removing the existing ones from revs:
         try {
+            // query db only, not necessary to be syncrhonized
```
pasin: This comment might not be correct if want to make calling to the database object thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1301,6 +1300,7 @@ public Status do_POST_Document_revs_diff(Database _db, String _docID, String _at
     public Status do_POST_Document_compact(Database _db, String _docID, String _attachmentName) {
         Status status = new Status(Status.OK);
         try {
+            // Make Database.compact() thread-safe
```
pasin: This comment might not be correct if want to make calling to the database object thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1563,6 +1564,7 @@ public Status do_GET_Document_changes(Database _db, String docID, String _attach
             Log.v(Log.TAG_ROUTER, "Filter params=" + changesFilterParams);
         }
 
+        // changesSince() is query only. not required synchronized
         RevisionList changes = db.changesSince(since, options, changesFilter, changesFilterParams);
```
pasin: This comment might not be correct if want to make calling to the database object thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1623,27 +1625,14 @@ public Status do_GET_Document(Database _db, String docID, String _attachmentName
                 String revID = getQuery("rev");  // often null
                 RevisionInternal rev = null;
                 if (isLocalDoc) {
+                    // query only -> not required synchronized
```
pasin: This comment might not be correct if want to make calling to the database object thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1623,27 +1625,14 @@ public Status do_GET_Document(Database _db, String docID, String _attachmentName
...
-                            int generation = RevisionInternal.generationFromRevID(ancestorId);
-                            db.stubOutAttachmentsIn(rev, generation + 1);
-                        }
-                    }
-                    */
                     boolean includeAttachments = options.contains(TDContentOptions.TDIncludeAttachments);
                     if (includeAttachments) {
                         options.remove(TDContentOptions.TDIncludeAttachments);
                     }
+                    // query only -> not required synchronized
```
pasin: This comment might not be correct if want to make calling to the database object thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1887,19 +1875,46 @@ private RevisionInternal update(Database _db, String docID, Body body, boolean d
...
+                try {
+                    fDb.getStore().runInTransaction(new TransactionalTask() {
+                        @Override
+                        public boolean run() {
+                            try {
+                                RevisionInternal r = fDb.getStore().putLocalRevision(_rev, _prevRevID, true);
+                                _revs.add(r);
+                                return true;
+                            } catch (CouchbaseLiteException e) {
+                                throw new RuntimeException(e);
```
pasin: This throw is not required as the runTransaction will catch any Exception and re-throw.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1984,6 +1999,7 @@ public Status do_PUT_Document(Database _db, String docID, String _attachmentName
                 throw new CouchbaseLiteException(Status.BAD_REQUEST);
             }
             List<String> history = Database.parseCouchDBRevisionHistory(body.getProperties());
+            // forceInsert uses transaction internally, not necessary to apply synchronized
             db.forceInsert(rev, history, source);
```
pasin: db.forceInsert is not thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1887,19 +1875,46 @@ private RevisionInternal update(Database _db, String docID, Body body, boolean d
...
+                    if (ex.getCause() != null &&
+                            ex.getCause().getCause() != null &&
+                            ex.getCause().getCause() instanceof CouchbaseLiteException)
+                        throw (CouchbaseLiteException) ex.getCause().getCause();
+                    else
+                        throw new CouchbaseLiteException(ex, Status.INTERNAL_SERVER_ERROR);
                 }
+            } else {
+                // putRevision() uses transaction internally.
+                result = _db.putRevision(rev, prevRevID, allowConflict);
```
pasin: This call here may or may not always thread safe.

https://github.com/couchbase/couchbase-lite-java-core/blob/2839fc7978e4ace5c67795c1df9b14653e65ae79/src/main/java/com/couchbase/lite/Database.java#L1750-L1757

src/main/java/com/couchbase/lite/Database.java
hideki: I have same concern.

src/main/java/com/couchbase/lite/router/Router.java
hideki: As [boolean run()](https://github.com/couchbase/couchbase-lite-java-core/blob/51b2bf2cccd0048fb2821cb527b315e3cc5e7102/src/main/java/com/couchbase/lite/TransactionalTask.java#L12) is not declared with throws Exception/CouchbaseLiteException. This is reason that re-throw exception with RuntimeException.

src/main/java/com/couchbase/lite/router/Router.java
hideki: [forceInsert()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1545) uses transaction in Store layer

https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1545

src/main/java/com/couchbase/lite/router/Router.java
hideki: [add()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1336) uses transaction in store layer.

src/main/java/com/couchbase/lite/router/Router.java
hideki: As [findMissingRevisions()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L963) does not access shard instance and SQLiteDatabae is thread-safe as default, I believe this method is thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
hideki: As [changesSince()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L963) does not access shard instance and SQLiteDatabae is thread-safe as default, I believe this method is thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
hideki: As [getLocalDocument()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1826) does not access shard instance and SQLiteDatabae is thread-safe as default, I believe this method is thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
hideki: As [getDocument()](https://github.com/couchbase/couchbase-lite-java-core/blob/a35361bc47ecb6c55849694deece22fa4aaf7c62/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L691) does not access shard instance and SQLiteDatabae is thread-safe as default, I believe this method is thread-safe.

src/main/java/com/couchbase/lite/router/Router.java
pasin: In theory, SQLiteDatabase is not (always) application logic thread-safe. SQLite is compiled with multi-threading mode, and Android SQLite database does have read/write connection separately. It's also related to connection pool size. We could chat more in detail.

src/main/java/com/couchbase/lite/router/Router.java
pasin: Please ignore my comment above.

src/main/java/com/couchbase/lite/router/Router.java
pasin: Same as above.

src/main/java/com/couchbase/lite/router/Router.java
pasin: Same as above.

src/main/java/com/couchbase/lite/router/Router.java
pasin: Same as above.

src/main/java/com/couchbase/lite/router/Router.java
pasin: Got your point.

src/main/java/com/couchbase/lite/router/Router.java
pasin: I see. Thanks.

src/main/java/com/couchbase/lite/router/Router.java
pasin: I see. Thanks.

src/main/java/com/couchbase/lite/router/Router.java
pasin: @hideki I took my word back. Android SQLiteDatabase seems to be thread-safe between read/write regardless of connection pool size. The reason behind this is related to the default number of connection pool size. By default the connection pool size is one and when enabling WAL mode by API not SQL command, connection pool size is 2 or 4 (I don't remember exactly).

```
WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.
```

We could chat more about what we should do :)

PR #1058, opened 2016-02-23
Fixed #1047 - Push replication skipped 20 documents

### LINKED ISSUE #1047: Push replication skipped 20 documents
Scenario: Device A is pushing and pulling with 4 other devices. So a total of 8 replications are running on Device A. When the user logs out from Device A, it stops the replications and deletes the database. While the user is logged out from Device A, 20 documents are added on each device (no active replications at this point). When the user logs back in Device A, all devices should have the same document count but it can sometimes happen that a device will be 20 documents behind in the total count. The replications are active again but this device will never receive the 20 documents that were skipped because the replicator has moved to tracking the last sequence number.

What's the most useful information I can provide to look into the root cause of this issue?

To begin, here's a thread dump of the device running the replications https://gist.github.com/jamiltz/df4b88c50e1e12ec3a05

hideki:
- Eliminate `synchronized(store)` from Database.java. They are not necessary, also it makes thread wait not right place.
- Make `SQLiteStore.compact()` thread safe. `PRAGMA wal_checkpoint(RESTART)` and `VACUUM` SQL query is suspicious.
- Make `runFilter()` out of SQL query loop to minimize holding SQLite connection in `changesSince()` method. `changesSince()` is frequently called for `/_changes` REST API. And it is bottle neck for listener.

### DIFF
--- a/src/main/java/com/couchbase/lite/store/SQLiteStore.java
+++ b/src/main/java/com/couchbase/lite/store/SQLiteStore.java
@@ -133,6 +133,7 @@ public class SQLiteStore implements Store, EncryptableStore {
     private StoreDelegate delegate;
     private int maxRevTreeDepth;
     private SymmetricKey encryptionKey;
+    private final Object compactLock = new Object(); // lock for compact() method
 
     ///////////////////////////////////////////////////////////////////////////
     // Constructor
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -628,39 +629,47 @@ public boolean inTransaction() {
...
+                    args.put("doc_type", (String) null);
+                    args.put("no_attachments", 1);
+                    int changes = storageEngine.update("revs", args, "current=0", null);
+                    Log.v(TAG, "... deleted %d revisions", changes);
+                } catch (SQLException e) {
+                    Log.e(TAG, "Error compacting", e);
+                    throw new CouchbaseLiteException(Status.INTERNAL_SERVER_ERROR);
+                }
+            } finally {
+                endTransaction(shouldCommit);
```
pasin: shouldCommit doesn't set to false in the catch block.

hideki:
@pasin,
I fixed endTransaction() issue. Please review this PR again? 
Thanks!

PR #1060, opened 2016-02-25
Fixed #1050 - reimplement push replicator pause logic and apply same waitForPendingFutures logic

### LINKED ISSUE #1050: Implement non-problematic flow control logic for push replicator
Current implementation of flow control logic for Push replicator could causes dead-lock. As `changed(ChangeEvent)` could be fired by `WorkExecutor` and it waits, so work executor hangs. Also `changed(ChangeEvent)` could be fired by UI thread, it might cause slow UI response.

Without flow control, push replicator could consume a lot of memory and it slows down replication.

NOTE: As simple solution, I thought about create new Thread in `changed()` callback and it waits if necessary. But this solution potentially crates so many threads and `ChangeEvent` order is not guaranteed.

Reference ticket: https://github.com/couchbase/couchbase-lite-java-core/issues/1047

hideki:
1. First commit: Previous commit had issue in case of adding changes from beginReplicating. Fixed the problem by not using ExecutorService.
1. Second commit: Applied same waitForPendingFutures logic to push replicator and simplified waitForPendingFutures method.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
@@ -48,6 +48,7 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
@@ -62,6 +63,8 @@ import java.util.concurrent.atomic.AtomicInteger;
  */
 @InterfaceAudience.Private
 abstract class ReplicationInternal implements BlockingQueueListener {
+
+    private static final String TAG = Log.TAG_SYNC;
 
     // Change listeners can be called back synchronously or asynchronously.
     protected enum ChangeListenerNotifyStyle {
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -1727,4 +1732,77 @@ protected boolean isRunning() {
...
+            waitPendingFuturesCompleted(pendingFutures);
+        }
+    }
+
+    protected static void waitBatcherCompleted(Batcher<RevisionInternal> b) {
+        // Wait for batcher completed
+        if (b != null) {
+            // if batcher delays task execution, need to wait same amount of time. (0.5 sec or 0 sec)
+            try {
+                Thread.sleep(b.getDelay());
```
pasin: @hideki I couldn't think of the reason why the sleep here is needed. Batcher.waitForpendingFutures() will wait until all tasks in the queues are done.

pasin:
@hideki I have one comment above but I will merge it now. Please review my comment. I believe you could get rid of the sleep.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
hideki: @pasin, I will review the this part of codes today. Thank you for your comment.

hideki:
TODO: https://github.com/couchbase/couchbase-lite-java-core/issues/1062

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
hideki: Without `sleep()`, `waitForPendingFutures()` exist before processing all items in batcher.

PR #1067, opened 2016-02-27
Fixed #1061 - NullPointerException in Implementation of Validator

### LINKED ISSUE #1061: NullPointerException in Implementation of Validator
Scenario : Implement a Validator that only log information about the revision that is about to be validated. Especially one must try to log the list of key that changed using the ValidationContext.getChangedKeys(). Once added to the database the validator will generate a NullPointerException when calling in ValidationContextImpl.java the following code : 
Map<String, Object> cur = getCurrentRevision().getProperties();
Also this NullPointerException gets thrown when calling isDeleted() from the document that is being validated in the current revision.

hideki:
In case of creating new document or deleting the document, Validator could cause NPE if access getKeyChanges() or isDeleted(). So added null check.

### DIFF
--- a/src/main/java/com/couchbase/lite/ValidationContextImpl.java
+++ b/src/main/java/com/couchbase/lite/ValidationContextImpl.java
@@ -42,16 +42,22 @@ class ValidationContextImpl implements ValidationContext {
     public List<String> getChangedKeys() {
         if (changedKeys == null) {
             changedKeys = new ArrayList<String>();
-            Map<String, Object> cur = getCurrentRevision().getProperties();
+            Map<String, Object> cur = null;
+            if (getCurrentRevision() != null)
+                cur = getCurrentRevision().getProperties();
             Map<String, Object> nuu = newRev.getProperties();
-            for (String key : cur.keySet()) {
-                if (!cur.get(key).equals(nuu.get(key)) && !key.equals("_rev")) {
-                    changedKeys.add(key);
+            if (null != null) {
+                if (cur != null) {
+                    for (String key : cur.keySet()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/ValidationContextImpl.java
```diff
@@ -42,16 +42,22 @@ public SavedRevision getCurrentRevision() {
...
             changedKeys = new ArrayList<String>();
-            Map<String, Object> cur = getCurrentRevision().getProperties();
+            Map<String, Object> cur = null;
+            if (getCurrentRevision() != null)
+                cur = getCurrentRevision().getProperties();
             Map<String, Object> nuu = newRev.getProperties();
-            for (String key : cur.keySet()) {
-                if (!cur.get(key).equals(nuu.get(key)) && !key.equals("_rev")) {
-                    changedKeys.add(key);
+            if (null != null) {
```
pasin: Should be nuu != null.

src/main/java/com/couchbase/lite/ValidationContextImpl.java
```diff
@@ -42,16 +42,22 @@ public SavedRevision getCurrentRevision() {
...
+                if (cur != null) {
+                    for (String key : cur.keySet()) {
+                        if (!cur.get(key).equals(nuu.get(key)) && !key.equals("_rev")) {
```
pasin: Would it be possible that cur.get(key) returns null?

hideki:
@pasin Please re-review this. Thank you very much for catching my mistake.

PR #1071, opened 2016-03-01
Fixed #1069 - Deleting database and stopping replications at the same…

### LINKED ISSUE #1069: Deleting database and stopping replications at the same time
Intermittent issue. I'm currently simultaneously deleting the database and stopping all replications at the same time and sometimes the call to DELETE /db doesn't return. However, making the same call 20/30 seconds later returns instantly.

Logs: https://gist.github.com/jamiltz/35ec4c60f18a27450f68

Search for "Delete DB [object Object]". It appears twice in the logs. First at `17:00:07` and then at `17:00:20`. The user clicked on logout button a second time 13 seconds later. The second time the button was clicked the DELETE /db request returned instantly.

hideki:
… time

- When calling `Database.delete() -> Database.close()`, some of replicators are stopping because of `/_replicate` with `cancel=true` requests. It seems some replicators stops before adding `Replication.ChangeLister`, then `latch.countDown()` for these replicators are never called. Then it makes following codes to wait 60 sec per missing STOPPED notification. In case of customer scenario, customer creates 10 replicators, worst case scenario is 10min waiting.
To avoid above, before calling `latch.await()`, check if replicator is still active. If yes, call `latch.await()`, otherwise, none.

Changes that are not related with this ticket:
- `databaseClosing()` method is same with `stop()` method. Eliminated it.
-  `readOnly` variable in Database, is never used. Eliminated it.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -91,7 +91,6 @@ public class Database implements StoreDelegate {
     private Store store = null;
     private String path;
     private String name;
-    private boolean readOnly = false;
     private boolean open = false;
 
     private Map<String, View> views;
@@ -141,7 +140,6 @@ public class Database implements StoreDelegate {
         this.path = path;
         this.name = name != null ? name : FileDirUtils.getDatabaseNameFromPath(path);
         this.manager = manager;
-        this.readOnly = readOnly;
 
         this.startTime = System.currentTimeMillis();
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1038,9 +1034,7 @@ public void removeDatabaseListener(DatabaseListener listener) {
     @InterfaceAudience.Private
     public List<Replication> getActiveReplications() {
         List<Replication> activeReplicatorsList = new ArrayList<Replication>();
-        if (activeReplicators != null) {
-            activeReplicatorsList.addAll(activeReplicators);
-        }
+        activeReplicatorsList.addAll(activeReplicators);
```
pasin: It's not clear to me if this operation is thread-safe even though activeReplicators is synchronized hash set. Can you confirm?

PR #1100, opened 2016-03-13
Fixed 1093 - Listener does not accept new connection

hideki:
- Make sure `InputStream.close()` is called whenever InputStream was created from `HttpEntity`
- Make sure `HttpEntity.consumeContent()` is called whenever `HttpClient.execute()` is called successfully. (Both HTTP OK and ERROR)
- Make sure `ConnectionManager.shutdown()` is called after HttpClient is created and is used.

Note: It looks many code changes. But they are indenting update.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
+++ b/src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
@@ -75,20 +75,21 @@ public class BulkDownloader extends RemoteRequest implements MultipartReaderDele
     @Override
     public void run() {
         HttpClient httpClient = clientFactory.getHttpClient();
-
-        preemptivelySetAuthCredentials(httpClient);
-
-        request.addHeader("Content-Type", "application/json");
-        request.addHeader("Accept", "multipart/related");
-        request.addHeader("User-Agent", Manager.getUserAgent());
-        request.addHeader("X-Accept-Part-Encoding", "gzip");
-        request.addHeader("Accept-Encoding", "gzip, deflate");
-
-        addRequestHeaders(request);
-
-        setBody(request);
-
-        executeRequest(httpClient, request);
+        try {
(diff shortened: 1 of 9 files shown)


src/main/java/com/couchbase/lite/replicator/BulkDownloader.java
```diff
@@ -120,29 +121,37 @@ protected void executeRequest(HttpClient httpClient, HttpUriRequest request) {
...
+                try {
+                    // conflict could be happen. So ERROR prints make developer confused.
+                    if (status.getStatusCode() == 409)
+                        Log.w(Log.TAG_REMOTE_REQUEST, "Got error status: %d for %s.  Reason: %s",
+                                status.getStatusCode(), request, status.getReasonPhrase());
+                    else
+                        Log.e(Log.TAG_REMOTE_REQUEST, "Got error status: %d for %s.  Reason: %s",
+                                status.getStatusCode(), request, status.getReasonPhrase());
+                    error = new HttpResponseException(status.getStatusCode(),
+                            status.getReasonPhrase());
```
pasin: There is a possibility that respondWithResult will be called twice on line#175 and 206.

src/main/java/com/couchbase/lite/support/RemoteMultipartDownloaderRequest.java
```diff
@@ -104,33 +117,28 @@ protected void executeRequest(HttpClient httpClient, HttpUriRequest request) {
...
                 }
                 finally{
-                    if(entity != null){try{ entity.consumeContent(); }catch (IOException e){}}
-                    entity = null;
+                    if (entity != null) {
+                        try {
+                            entity.consumeContent();
+                        } catch (IOException e) {
+                        }
+                    }
```
pasin: Don't we need to send respondWithResult for the outer Exception (error case)?

PR #1166, opened 2016-03-27
FIXED #1147 - [Scenario 4] Missing changes in Database Change Notific…

### LINKED ISSUE #1147: [Scenario 4] Missing changes in Database Change Notification
I understand that the Database.postChangeNotifications() could be called from different threads. Two variables in the method are not thread safe, `postingChangeNotifications` and `changesToNotify`.

hideki:
…ation

As @pasin analyzed, `changesToNotify`, `postingChangeNotifications`, `changeListeners`, and `databaseListeners` variables are not thread-safe. In this scenario, `changeListener` and `databaseListners` are not added/removed in concurrent env. These two variables don't directly cause the problem with under unit-test scenario. `changesToNotify` and `postingChangeNotifications` makes test fails. Especially, `changesToNotify`  variable. By this commit, make all four variables thread-safe. And make `postChangeNotifications()` method thread-safe by `postingChangeNotifications` variable and `lockPostingChangeNotifications` lock object.

I run `testDatabaseChangeNotification()` unit test with this fix many times with Genymotion emulator (faster test env) and Stock Emulator ARM API 19 (slower test env). With this fix, the test never failed again. I also all unit test with this fix with both genymotion and ARM API 19 emulator. And test passed.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -107,7 +107,7 @@ public class Database implements StoreDelegate {
     private Cache<String, Document> docCache;
     final private List<DocumentChange> changesToNotify;
     private boolean postingChangeNotifications;
-    final protected Object lockPostingChangeNotifications = new Object();
+    final private Object lockPostingChangeNotifications = new Object();
     private long startTime;
 
     /**
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -58,7 +58,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -103,11 +102,12 @@
...
     private Manager manager;
-    final private CopyOnWriteArrayList<ChangeListener> changeListeners;
-    final private CopyOnWriteArrayList<DatabaseListener> databaseListeners;
+    final private Set<ChangeListener> changeListeners;
+    final private Set<DatabaseListener> databaseListeners;
     private Cache<String, Document> docCache;
-    private List<DocumentChange> changesToNotify;
+    final private List<DocumentChange> changesToNotify;
     private boolean postingChangeNotifications;
+    final protected Object lockPostingChangeNotifications = new Object();
```
pasin: Why do we need protected?

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -2194,48 +2196,55 @@ protected void rememberAttachmentWriter(BlobStoreWriter writer) {
...
                 }
-
-                posted = true;
-            } catch (Exception e) {
-                Log.e(Database.TAG, this + " got exception posting change notifications", e);
-            } finally {
+            }
+            return posted;
+        } finally {
+            synchronized (lockPostingChangeNotifications) {
```
pasin: Can we do `synchronized (lockPostingChangeNotifications)` before `return posted` so we don't need an extra finally block?

src/main/java/com/couchbase/lite/Database.java
hideki: It should be `private`. Thank you.

src/main/java/com/couchbase/lite/Database.java
hideki: If `postingChangeNotifications` variable is not set to `false`, `postChangeNotifications()` method can not be execute any more. Between Line 2206 and Line 2241, the method does not return, and it does not throw an Exception. From code point of view, using `finally` block is not necessary. However there is possibility methods that `postChangeNotifications()` method calls could throw RuntimeException. So this is reason, I use finally block to reset `postingChangeNotifications` variable. What do you think?

pasin: RuntimeException is a subclass of the Exception so the catch statement on line 2238 should be able to catch it.

pasin: Unless you mean, `while (!store.inTransaction() && isOpen() && changesToNotify.size() > 0)` might throw an exception.

PR #1181, opened 2016-04-01
Fixed #1180 - Router: PUT /{db}/{doc} If-Match header is not supported

### LINKED ISSUE #1180: Router: PUT /{db}/{doc} If-Match header is not supported
See spec:
http://developer.couchbase.com/documentation/mobile/1.2/develop/references/couchbase-lite/rest-api/document/put---db---doc-/index.html

hideki:


src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -375,6 +375,14 @@ private static String getAccept(URLConnection connection) {
...
     }
 
+    private static String getIfMatch(URLConnection connection) {
+        String ifMatch = connection.getRequestProperty("If-Match");
+        if (ifMatch == null)
+            // From Android: http://developer.android.com/reference/java/net/URLConnection.html
+            ifMatch = connection.getRequestProperty("if-match");
+        return ifMatch;
+    }
+
```
pasin: I have noticed that you started to wrap connection.getRequestProperty() call with normal case and lowercase property name. I have seen a lot of connection.getRequestProperty() in Router.java. Do all connection.getRequestProperty() calls need a wrapper like this? If YES, it is better to have one method to do this instead of one per property.

hideki:
@pasin,
I applied your suggestions to PR.
Could you please review this again?
Thanks!

PR #1197, opened 2016-04-13
Fixed Android Issue 834 - ConcurrentModificationException when try to…

hideki:
… update object

- Ticket: https://github.com/couchbase/couchbase-lite-android/issues/834

- Use `CopyOnWriteArraySet` or `CopyOnWriteArrayList` for Set/List of changeListener. CopyOnWriteArraySet/CopyOnWriteArrayList creates copy of array for iteration, so it does not cause `ConcurrentModificationException` and it is thread-safe.
- Except `changeListeners`, not applied CopyOnWriteArraySet/CopyOnWriteArrayList
- In addition, applied `final` for variables that should not be changed.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -88,9 +88,9 @@ public class Database implements StoreDelegate {
     public static int DEFAULT_MAX_REVS = 20;
 
     private Store store = null;
-    final private String path;
+    private final String path;
     private String name;
-    final private AtomicBoolean open = new AtomicBoolean(false);
+    private final AtomicBoolean open = new AtomicBoolean(false);
 
     private Map<String, View> views;
     private Map<String, String> viewDocTypes;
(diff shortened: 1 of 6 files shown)


src/main/java/com/couchbase/lite/Database.java
```diff
@@ -87,7 +88,7 @@
     public static int DEFAULT_MAX_REVS = 20;
 
     private Store store = null;
-    private String path;
+    final private String path;
```
pasin: I think in the other classes, we use `private final`. It's better to have one style.

src/main/java/com/couchbase/lite/Document.java
```diff
@@ -33,7 +34,7 @@
     /**
      * Change Listeners
      */
-    private List<ChangeListener> changeListeners = Collections.synchronizedList(new ArrayList<ChangeListener>());
+    final private List<ChangeListener> changeListeners = new CopyOnWriteArrayList<ChangeListener>();
```
pasin: Same here. Using `private final` is more common, I guessed.

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -77,7 +78,7 @@
     protected ScheduledExecutorService workExecutor;
     protected ReplicationInternal replicationInternal;
     protected Lifecycle lifecycle;
-    protected List<ChangeListener> changeListeners;
+    final private List<ChangeListener> changeListeners = new CopyOnWriteArrayList<ChangeListener>();
```
pasin: Same as above.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -114,7 +115,7 @@
     protected ScheduledExecutorService workExecutor;
 
     protected StateMachine<ReplicationState, ReplicationTrigger> stateMachine;
-    protected List<ChangeListener> changeListeners;
+    final private List<ChangeListener> changeListeners = new CopyOnWriteArrayList<ChangeListener>();
```
pasin: Same as above.

hideki:
@pasin Please review this PR again.

PR #1203, opened 2016-04-15
Fixed #1194 - Fallback to GET when POST is not allowed to _changes

### LINKED ISSUE #1194: Fallback to GET when POST is not allowed to _changes
CBLSocketChangeTracker.m
```objective-c
- (void) failedWithError:(NSError*) error {
    [self clearConnection];

    // Work around Cloudant's lack of support of POST to _changes. a 405 (Method Not Allowed) is
    // a dead giveaway, but Cloudant returns a 401 instead if the database is write-protected.
    // This could also happen with a read-protected SG or CouchDB database, so if we get a 401
    // double-check that it's Cloudant. (See issue #1020.)
    if (_usePOST) {
        if ([error my_hasDomain: CBLHTTPErrorDomain code: kCBLStatusMethodNotAllowed]
            || ([error my_hasDomain: NSURLErrorDomain code: NSURLErrorUserAuthenticationRequired]
                    && [_serverName rangeOfString: @"CouchDB/1.0.2"].length > 0)) {
                LogTo(ChangeTracker, @"Apparently server is Cloudant; retrying with a GET...");
                _usePOST = NO;
                _http = nil;
                [self retryAfterDelay: 0.0];
                return;
            }
    }

    [super failedWithError: error];
}
```

- Also need to review `setUsePOST(true)`

hideki:
- If https status code is 405 for POST /{db}/_changes, retry it with GET.
- Applied fhttps://github.com/couchbase/couchbase-lite-ios/pull/1187 - Fixed incompatibility with CouchDB's _changes feed

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
+++ b/src/main/java/com/couchbase/lite/replicator/ChangeTracker.java
@@ -660,7 +660,7 @@ public class ChangeTracker implements Runnable {
         post.put("feed", getFeed());
         post.put("heartbeat", getHeartbeatMilliseconds());
         post.put("style", includeConflicts ? "all_docs" : null);
-        post.put("active_only", activeOnly && !caughtUp ? "true" : null);
+        post.put("active_only", activeOnly && !caughtUp ? true : null);
         post.put("since", since);
         post.put("limit", limit > 0?limit:null);
         // TODO: {@"accept_encoding", @"gzip"}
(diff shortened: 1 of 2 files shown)


codingdudecom:
hey guys, I've started my first project with couchbase lite on android 2 weeks ago, and went insane as I thought I was doing something wrong with the views. luckily enough you fixed it and now it works for me. thanks a million!

PR #1230, opened 2016-05-09
Fixed Android Issue 863 - LiveQuery is blocked by replicator

hideki:
This commit is second step to fix issue 863: replicator own own executor service instead of using workExecutor.

Area needs your attention: 
- With this commit, the executor is shutdown in `finalize()` of `ReplicationInternal`. I am not 100% sure if it is ok. Ideally `ExecuterService.shutdown()` should be called out-side of GC.
- If app keeps creating new Replicator, thread leak is caused till app close Database because database holds all replicators which also hold ReplicationInternal.


pasin:
I don't have any comments. Can I merge this one before the other one?

hideki:
yes

pasin:
Can you fix the conflict?

hideki:
Fixed conflict.

PR #1291, opened 2016-06-24
Fixed #1285 : Change HttpClientFactory to create only one shared OkHttpClient

### LINKED ISSUE #1285: OkHttp ConnectionPool thread leaking
It seems like OkHttp ConnectionPool threads keep growing, while the replication is running. I have noticed this while working on #1275.

pasin:
Instead of creating a new OkHttpClient object every time when getting the http client, create only one shared object. This will result to one OkHttpClient per replicator and avoid to have connection pool thread leak.

#1285

### DIFF
--- a/src/main/java/com/couchbase/lite/support/CouchbaseLiteHttpClientFactory.java
+++ b/src/main/java/com/couchbase/lite/support/CouchbaseLiteHttpClientFactory.java
@@ -32,7 +32,7 @@ import okhttp3.CookieJar;
 import okhttp3.OkHttpClient;
 
 public class CouchbaseLiteHttpClientFactory implements HttpClientFactory {
-
+    private OkHttpClient client;
     private ClearableCookieJar cookieJar;
     private SSLSocketFactory sslSocketFactory;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/support/CouchbaseLiteHttpClientFactory.java
```diff
@@ -32,7 +32,7 @@
 import okhttp3.OkHttpClient;
 
 public class CouchbaseLiteHttpClientFactory implements HttpClientFactory {
-
+    private OkHttpClient client;
```
hideki: Is holding OkHttpClient instance in Factory best place?

pasin: I think so.

pasin: Given the fact that each replicator creates a new factory and passes the factory around, without hugh refactory, keeping inside the factory is the best place.

Even without the fact above, keep it inside the factory seem to be a good idea. It is easier and less messy to keep the object near its source.

hideki: Following is my personal preference. I understand that your solution is effective with minimum changes. For 1.3 release, this is good enough. But I would like to refactor after we release 1.3.

- In general, Factory is to create a target instance. If holding instance in a class, I prefer different name rather than Factory.
- By holding one instance in a Factory, it guarantees one instance per a factory. But this does not guarantees one instance per a replicator or a database. I prefer holding OkHttpClient instance in a replicator or a database.

hideki: I created the ticket for refactoring HttpClientFactory.
https://github.com/couchbase/couchbase-lite-java-core/issues/1292

pasin: Agree with all your points.

PR #1305, opened 2016-06-28
Fixed https://github.com/couchbase/couchbase-lite-android/issues/898

hideki:
Push replication failing with large data sets

- Solution: Implement pause logic into PusherInternal. In case total count of Batcher and pendingSequences sizes are more than 200, push replicator pauses to add changes into Batcher. This is same approach with PullerInternal pause logic.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
@@ -220,15 +220,6 @@ public class PusherInternal extends ReplicationInternal implements Database.Chan
     @Override
     @InterfaceAudience.Private
     public void beginReplicating() {
-        supportExecutor.submit(new Runnable() {
-            @Override
-            public void run() {
-                beginReplicatingInternal();
-            }
-        });
-    }
-
-    private void beginReplicatingInternal() {
         // If we're still waiting to create the remote db, do nothing now. (This method will be
         // re-invoked after that request finishes; see -maybeCreateRemoteDB above.)
 
--- a/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
@@ -72,10 +72,10 @@ public class PusherInternal extends ReplicationInternal implements Database.Chan
     boolean doneBeginReplicating = false;
     List<RevisionInternal> queueChanges = new ArrayList<RevisionInternal>();
     private String str = null; // for toString()
-
+    // for pause
     private boolean paused = false;
     private final Object pausedObj = new Object();
-    private ExecutorService supportExecutor;
+    private ExecutorService supportExecutor; // executor to submit revision into batcher
 
     /**
      * Constructor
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/replicator/PusherInternal.java
```diff
@@ -209,23 +271,25 @@ public void beginReplicating() {
...
             //       submit changes  into inbox, there are chance both inbox and pendingFutures are
             //       empty.
             submitRevisions(changes);
         } else {
-            Log.d(Log.TAG_SYNC, "%s: No changes since %d", this, lastSequenceLong);
+            Log.d(TAG, "%s: No changes since %d", this, lastSequenceLong);
         }
 
         // process queued changes by `changed()` callback
         synchronized (changesLock) {
```
pasin: I think synchronize here doesn't need anymore as the code will be run in the support work executor.

src/main/java/com/couchbase/lite/replicator/PusherInternal.java
```diff
@@ -209,23 +271,25 @@ public void beginReplicating() {
...
+        Log.d(TAG, "%s: Getting changes since %d", this, lastSequenceLong);
         final RevisionList changes = db.changesSince(lastSequenceLong, options, filter, filterParams);
         if (changes.size() > 0) {
-            Log.d(Log.TAG_SYNC, "%s: Queuing %d changes since %d", this, changes.size(), lastSequenceLong);
+            Log.d(TAG, "%s: Queuing %d changes since %d", this, changes.size(), lastSequenceLong);
             // NOTE: Needs to submit changes into inbox from RemoteRequest thread for beginReplication.
             //       RemoteRequest thread is observed by pendingFuture, if using other thread to
             //       submit changes  into inbox, there are chance both inbox and pendingFutures are
             //       empty.
             submitRevisions(changes);
```
pasin: Does the NOTE above this line is still applicable?

src/main/java/com/couchbase/lite/replicator/PusherInternal.java
```diff
@@ -98,14 +115,48 @@ public void setCreateTarget(boolean createTarget) {
...
-        Log.d(Log.TAG_SYNC, "%s STOPPING...", toString());
+        Log.d(TAG, "%s STOPPING...", toString());
 
         stopObserving();
 
+        // Awake thread if it is wait for pause
+        setPaused(false);
+
+        // shutdown supportExecutor immediately, does not add any more tasks.
+        terminateSupportExecutor();
```
pasin: Could this cause race condition between terminateSupportExecutor() and adding tasks to the executor? The race condition might be ok if adding the tasks to the work executor after being terminated has no side effect.

hideki: As you wrote, I believe race-condition does not cause the problem. Once executor is shutdown, any new tasks are not processed.

src/main/java/com/couchbase/lite/replicator/PusherInternal.java
hideki: Thank you for catching this. This is still applicable.

hideki:
@pasin, Could you please review again?

Changes
- make `beginReplicating()` not call from `ExecutorService` to avoid potential problem
- Instead, make minimum thread-pool size of remote request executor to 2

pasin:
We should create a ticket to rethink about the replicator work executor and the way that waits for pending future and etc.

PR #1330, opened 2016-07-06
Fixed #1319 - Auto-pruning failing on pull replication

### LINKED ISSUE #1319: Autopruning failing on pull replication
Step to reproduce:
Using `Android 1.3.0-8`

1. Start LiteServ
2. Start Sync Gateway
3. Add 1 doc to Sync Gateway
5. Update Doc 100 times on SG
7. Set Up Continuous Pull from SG
```
POST http://<liteserv_url>:5984/_replicate
HEADERS = {'Content-Length': '81', 'Content-Type': 'application/json'}
BODY = {"source": "http://<sg_url>:4985/db", "continuous": true, "target": "ls_db"}	
```
```
{"session_id":"repl001"}
```
8. Wait for replications to finish

Query doc with `revs=true`. 

Expected: Client has 20 revisions
Actual: Client has more than 20 revisions

hideki:
- original logic could not handle the case of pull replication with batch.

### DIFF
--- a/src/main/java/com/couchbase/lite/store/SQLiteStore.java
+++ b/src/main/java/com/couchbase/lite/store/SQLiteStore.java
@@ -1751,13 +1751,19 @@ public class SQLiteStore implements Store, EncryptableStore {
 
             // Delete the deepest revs in the tree to enforce the maxRevTreeDepth:
             int gen = inRev.getGeneration();
-            if (localRevs != null && gen > maxRevTreeDepth) {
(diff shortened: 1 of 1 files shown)


pasin:
@borrrden Not sure when .NET implemented this feature. You may double check if .NET has the same issue or not.

borrrden:
@pasin I ran into a similar issue before during functional testing and already corrected it, thanks!

Issue #1315, opened 2016-06-30
Replicator shouldn't go to IDLE state if the authenticating is not done


pasin:
This found when working on https://github.com/couchbaselabs/GrocerySync-Android/tree/feature/openid for OIDC testing.

PR #1336, opened 2016-07-07
Fixed #1333 - Obtaining username with refresh-token might have issue

### LINKED ISSUE #1333: OIDC: Obtaining username with refresh-token might have issue
In case of restarting application (GrocerySync), need to obtain username by using refresh-token. But it seems this logic does not work correctly.

hideki:
- during authentication, not send Replicator State Transition notification.


pasin:
This PR also contains #1315 fix. Can you fix this?

PR #1351, opened 2016-07-14
Fixed https://github.com/couchbase/couchbase-lite-android/issues/884

hideki:
[CouchDB 1.6.1] ver. 1.2.1 has Exception uploading multipart request org.apache.http.client.ClientProtocolException

- Problem: OkHttp use chunked transfer for multipart document request if content-length is unknown. And CouchDB 1.6.1 disconnects connection for it.
- Solution: Needs to return actual content-length if server is CouchDB. SyncGateway can handle chunked transfer.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/BlobRequestBody.java
+++ b/src/main/java/com/couchbase/lite/replicator/BlobRequestBody.java
@@ -1,16 +1,16 @@
-/**
- * Copyright (c) 2016 Couchbase, Inc. All rights reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
- * except in compliance with the License. You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software distributed under the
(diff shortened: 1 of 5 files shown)


src/main/java/com/couchbase/lite/replicator/BlobRequestBody.java
```diff
@@ -42,6 +44,36 @@ public MediaType contentType() {
...
+             * To send non-chunked transfer with multipart document by OkHttp, the client needs
+             * to specify content-length.
+             * Sync Gateway can accept chunked transfer
+             */
+            @Override
+            public long contentLength() throws IOException {
+                // Sync Gateway
+                if (syncGateway) {
+                    return super.contentLength();
+                }
```
pasin: Is chunked transfer the preferred method by SyncGateway? Can we have only one logic?

hideki: I am not sure if chunked transfer is preferred method for Sync Gateway. But Sync Gateway can handle chunked transfer, and we tested it for a while. So I consider it is not good time to switch.

I also talked with @adamcfraser. Sync Gateway does not care if it is chunked transfer or not.

PR #1368, opened 2016-07-25
Fixed https://github.com/couchbase/couchbase-lite-android/issues/929

hideki:
SQLiteDatabaseLockedException in 1.3.0-8

see: https://github.com/couchbase/couchbase-lite-android/issues/929#issuecomment-234975245


src/main/java/com/couchbase/lite/store/SQLiteStore.java
```diff
@@ -63,6 +64,10 @@
 
     public static String kDBFilename = "db.sqlite3";
 
+    private static final int kSQLiteBusyTimeout = 5000; // 5sec - Currently not used - default: 2.5sec
```
pasin: Can we remove this as it's not getting used?

PR #1384, opened 2016-08-14
Fixed #1296 : Implement GET _changes feed timeout to prevent dead connection leaks

### LINKED ISSUE #1296: Cordova - After putting Android app in background mode for a few hours, CB lite is unresponsive
Issue was originally reported at forums.couchbase.com
https://forums.couchbase.com/t/cordova-after-putting-android-app-in-background-mode-for-a-few-hours-cb-lite-is-unresponsive/8833

```
Hey everyone,

As stated in the subject line, my Cordova app running on Android is experiencing issues with connecting to CB lite after the app has been put in background mode for a number of hours. This is happening to all of my Android users and me.

When I look at the network log of the traffic of the app after resumption, all CB lite calls are pending.

I need urgent help on this as it's killing user adoption on Android. I'm not experiencing this issue on IOS.

Is there a inactivity timeout setting I can adjust?
```

pasin:
Implemented GET _changes feed timeout to prevent dead connection leaks. The implementation is inline with couchdb and SG. The default timeout is 60000 miliseconds (same as couchdb but different from SG which has a default timeout as 5 mins) and the timeout will be disabled if heartbeat is specified.

Note that couchbase-lite-ios doesn’t implement the timeout feature as the iOS listener doesn’t seem to have half-close socket issues as couchbase-lite-android is having.

#1296

### DIFF
--- a/src/main/java/com/couchbase/lite/router/Router.java
+++ b/src/main/java/com/couchbase/lite/router/Router.java
@@ -51,6 +51,7 @@ import com.couchbase.lite.support.Version;
 import com.couchbase.lite.util.Log;
 import com.couchbase.lite.util.StreamUtils;
 import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/router/Router.java
```diff
@@ -1595,91 +1609,141 @@ private void sendContinuousChange(RevisionInternal rev) {
...
+        try {
+            Map<String, Object> body = responseBodyForChanges(revs, since);
+            if (callbackBlock != null) {
+                byte[] data = null;
+                try {
+                    data = Manager.getObjectMapper().writeValueAsBytes(body);
+                } catch (Exception e) {
+                    Log.w(TAG, "Error serializing JSON", e);
+                }
+                os.write(data);
```
hideki: In case `writeValueAsBytes(body)` throws exception, `data` is null. Is `os.write(null)` OK?

pasin: I agree that the result from the code could be unexpected. This code is inherited from the current code. I don't want to change the behavior as we don't see any issues out of this.

Issue #939, opened 2016-01-22
Optimize first-time sync


zgramana:
See https://github.com/couchbase/sync_gateway/issues/1503

hideki:
iOS implementation: https://github.com/couchbase/couchbase-lite-ios/commit/eef3f86b0255fd852e45a3828a4784cf659da89c

PR #1388, opened 2016-08-17
Fixed #1386 and andriod #939 : POST /_replicate to start multiple replicators and threading issues

### LINKED ISSUE #1386: POST _replicate not create replicators when altering props beside type, target/source
1. Send request to <ListenerURL>/_replicate with the following body:
 ```
 {
      source: 'db',
      target: 'http://<SGW URL>:4984/db',
      continuous: true
 }
 ```

2. Send another request to <ListenerURL>/_replicate with the following body:
 ```
 {
      source: 'db',
      target: 'http://<SGW URL>:4984/db',
      continuous: true,
      doc_ids: ['doc1]
 }
 ```

3. Observe from the listener logcat and thread dump and found that there is only one replicator created.

4. The result from step 1 and Step 2 has the same session id which are not correct.

5. GET /_active_tasks returns only 1 task instead of 2 tasks.

pasin:
- Fixed creating multiple replicators of the same properties from sending
   multiple simultaneously POST /_replicate requests. The fix is synchronizing
   manager object in Router.do_POST_replicate() method.
   Ref: https://github.com/couchbase/couchbase-lite-android/issues/939

- Fixed not being able to create different replicators that have the same URL
   and type but different other properties. The fix is to ensure that
   Manager.getReplicator() searches a new create replicator against
   the current active replicators correctly. Ref: #1386

- [PLUS] In Router, extracted send response code into its own method
   called `sendResponseCodeAndFinish()` to avoid writing this same
   piece of codes several time.

- [PLUS to Router] Added CORS preflight /OPTIONS requests support to
   allow to test the listener with Web App.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -464,7 +464,8 @@ public class Replication
     /**
      * Following two methods for temporary methods instead of CBL_ReplicatorSettings implementation.
      */
-    protected String remoteCheckpointDocID() {
+    @InterfaceAudience.Private
+    public String remoteCheckpointDocID() {
         return replicationInternal.remoteCheckpointDocID();
     }
 
@@ -937,20 +938,5 @@ public class Replication
     @Override
     public String toString() {
         return "Replication{" + remote + ", " + (isPull() ? "pull" : "push") + '}';
-    }
-
-    @Override
-    public int hashCode() {
-        return replicationInternal.remoteCheckpointDocID().hashCode();
-    }
-
-    @Override
-    public boolean equals(Object obj) {
-        if (obj instanceof Replication) {
-            Replication replication = (Replication)obj;
-            return this.replicationInternal.remoteCheckpointDocID().equals(
-                    replication.replicationInternal.remoteCheckpointDocID());
-        }
-        return false;
     }
 }
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -938,4 +938,19 @@ public class Replication
     public String toString() {
         return "Replication{" + remote + ", " + (isPull() ? "pull" : "push") + '}';
     }
+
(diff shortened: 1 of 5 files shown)


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -938,4 +938,19 @@ private void addProperties(ReplicationInternal replicationInternal) {
...
         return "Replication{" + remote + ", " + (isPull() ? "pull" : "push") + '}';
     }
+
+    @Override
+    public int hashCode() {
+        return replicationInternal.remoteCheckpointDocID().hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
hideki: Please add.
if ( this == obj ) return true;

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -938,4 +938,19 @@ private void addProperties(ReplicationInternal replicationInternal) {
     public String toString() {
         return "Replication{" + remote + ", " + (isPull() ? "pull" : "push") + '}';
     }
+
+    @Override
+    public int hashCode() {
+        return replicationInternal.remoteCheckpointDocID().hashCode();
```
hideki: remoteCheckpointDocID() could return `null`

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -938,4 +938,19 @@ private void addProperties(ReplicationInternal replicationInternal) {
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof Replication) {
+            Replication replication = (Replication)obj;
+            return this.replicationInternal.remoteCheckpointDocID().equals(
+                    replication.replicationInternal.remoteCheckpointDocID());
```
hideki: remoteCheckpointDocID() could return `null`

pasin:
I push the fix per your review. Could you please re-review?

PR #1390, opened 2016-08-17
fixed bug in getPendingDocIDs

terrydhariwal:
Fixed bug in getPendingDocIDs to reflect the correct number of pending documents while Lite is offline.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -470,8 +470,18 @@ public class Replication
 
     public Set<String> getPendingDocumentIDs() {
         synchronized (lockPendingDocIDs) {
-            if (isPull() || (isRunning() && pendingDocIDs != null))
-                return pendingDocIDs;
+
+            /** 
+             * Bug causing code                
+             * Expected Functionality:
+             *      While offline, I expect the return value of this method to reflect
+             *      the number of pendingDocIDs inline with database mutations 
+             * Actual Behavior:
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -470,8 +470,18 @@ protected String remoteCheckpointDocID() {
...
+             * Bug causing code                
+             * Expected Functionality:
+             *      While offline, I expect the return value of this method to reflect
+             *      the number of pendingDocIDs inline with database mutations 
+             * Actual Behavior:
+             *      The commented out lines below, output a cached value for pendingDocIDs 
+             *
+             * if (isPull() || (isRunning() && pendingDocIDs != null))
+             *   return pendingDocIDs;
+            */
```
pasin: Removing this check entirely is not correct. First this method should return null for pull replicator.

It's best to port the changes from CBL iOS: 

https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/API/CBLReplication.m#L445-L451

terrydhariwal: Hi, I can leave in the null check for pull replicator. 

However, once we've determined that we're dealing with a push, we should proceed without returning the cached value of pendingDocIDs (since that will not reflect new updates to lite)

Do you agree ?

hideki:
Hi @terrydhariwal ,
Thank you for PR.
We would fix the issue correctly. I will close this PR.
Can you file the ticket to https://github.com/couchbase/couchbase-lite-java-core/issues with detailed information?
Thanks!
hideki

PR #1395, opened 2016-08-21
updated getPendingDocumentIDs() to return non-cached version while offline

terrydhariwal:
This fixes a bug where the pendingDocIDs returned is incorrect. I've maintained a check for isPull as requested. I know you wish to port changed from CBL iOS (https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/API/CBLReplication.m#L445-L451) but can we at least have this change accepted in the meantime because it does fix the issue

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -483,6 +483,9 @@ public class Replication
              *   return pendingDocIDs;
             */
 
+            if (isPull())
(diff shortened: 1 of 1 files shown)


terrydhariwal:
This is relation to bug https://github.com/couchbase/couchbase-lite-java-core/issues/1396

hideki:
@terrydhariwal Thank you for your PR. However, I'd like to fix this different way. Thanks!

PR #1402, opened 2016-08-25
Fixed https://github.com/couchbase/couchbase-lite-android/issues/888

hideki:
Issue: Crash during 1.1 to 1.2 migration causes permanent loss of data

Changes:
- Instead of migrate to actual new name, create temporary database and migrate to temporary db, then rename temporary name to actual name.
- If temporary database exists, rollback it to old name before continue to migration. Rollback actually rollback attachment files.
- Because of calling backOut() for rollback, importData() method do move file attachment instead of copy file. <- No change.

### DIFF
--- a/src/main/java/com/couchbase/lite/Manager.java
+++ b/src/main/java/com/couchbase/lite/Manager.java
@@ -821,7 +821,8 @@ public final class Manager {
             // Create and open new CBLDatabase with temporary name:
             Database tmpDB = getDatabase(tempName, false);
             if (tmpDB == null) {
-                Log.w(Log.TAG_DATABASE, "Upgrade failed: Creating new db failed: %s", tempName);
+                Log.w(Log.TAG_DATABASE,
+                        "Upgrade failed: Creating new db failed: %s", tempName);
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


pasin:
Do we also need to handle the current broken database upgrade that user is encountering?

hideki:
If user does not use the app after crash during db upgrade, recovering it is useful. But If user has kept using app after data loss, recovering could cause another data loss. Because of this, I did not add the code to handle current broken database. What do you think?

pasin:
Actually I don't know the best way to handle this case. Based on the issue (https://github.com/couchbase/couchbase-lite-android/issues/888), it seems like users want to recover the data.

pasin:
@snej could you do a second review and provide your opinion about handling the current broken database after database update crash?

src/main/java/com/couchbase/lite/Manager.java
```diff
@@ -804,28 +805,81 @@ public boolean accept(File file, String name) {
...
+            }
+
+            // close temporary database
+            tmpDB.close();
+
+            // rename temporary database name to new name
+            File tmpPath = new File(pathForDatabaseNamed(tempName));
+            File newPath = new File(pathForDatabaseNamed(name));
+            if (!tmpPath.renameTo(newPath)) {
+                Log.w(Log.TAG_DATABASE, "Upgrade failed: Failed to rename db folder from temporary name: %s -> %s", tmpPath, newPath);
```
pasin: I think we also need to backOut the change here as well.

PR #1420, opened 2016-09-01
Fixed #1419 : Ensure cursor is closed in finally block

### LINKED ISSUE #1419: Cursor not getting closed in finally block
While I reviewed the code for https://github.com/couchbase/couchbase-lite-android/issues/985, I have found this. If an exception happens and is handled, the connection will never been returned to the pool.

https://github.com/couchbase/couchbase-lite-java-core/blob/af4d13b089b7885788950ec95d5f4d698493bb43/src/main/java/com/couchbase/lite/store/SQLiteStore.java#L1318-L1319

The fix is easy. Just move those 2 lines inside the try block.

pasin:
Make sure that the cusor is created and used in the try/catch loop and is closed in the finally loop.

#1419

### DIFF
--- a/src/main/java/com/couchbase/lite/store/SQLiteStore.java
+++ b/src/main/java/com/couchbase/lite/store/SQLiteStore.java
@@ -749,7 +749,6 @@ public class SQLiteStore implements Store, EncryptableStore {
 
         Cursor cursor = null;
         try {
-            cursor = null;
             String cols = "revid, deleted, sequence";
             if (withBody) {
                 cols += ", json";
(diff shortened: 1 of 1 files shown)


hideki:
If we merge this into master, this fix could be in 1.3.1 because we need to build final build from master. If this fix is for 1.4.0 or potentially in 1.3.1, can you create dev branch and create PR against dev branch?

pasin:
I think this can be merged after 1.3.1.

hideki:
Note: merge after 1.3.1 release

PR #1422, opened 2016-09-01
Fixed #1421 Allow lastError to be cleared from outside

### LINKED ISSUE #1421: Replicator.lastError is not reset when restarting replicator.
I'm using the following pattern to monitor session expiration and other replication error.
```
private Replication.ChangeListener getReplicationChangeListener() {
        return new Replication.ChangeListener() {

            @Override
            public void changed(Replication.ChangeEvent event) {
                Replication replication = event.getSource();
                if (event.getError() != null) {
                    Throwable lastError = event.getError();
                    if(lastError.getMessage()!=null) {
                        Log.e(TAG, lastError.getMessage());
                        if (lastError instanceof RemoteRequestResponseException) {
                            RemoteRequestResponseException responseException = (RemoteRequestResponseException) lastError;
                            Map info = ((RemoteRequestResponseException) lastError).getUserInfo();
                            if (responseException.getCode() == 401 && replication.isPull() && info==null) {
                                onSyncUnauthorizedObservable.notifyChanges();
                                event.getSource().setLastError(null); //or .clearLastError()
                            }
                        }
                    }
                }
                EventBus.getDefault().post(event);
            }
        };
    }
```
Seems that after an authentication failure (error 401), in the next change events, even in the succesful ones, getError() is still returning that previously handled error.
So, I want to clear the error with setLastError(null), after the error is dealt with.
Is it possible to make Replication.setLastError() public?
https://github.com/couchbase/couchbase-lite-java-core/blob/master/src/main/java/com/couchbase/lite/replicator/Replication.java#L457-L463

I'm using
cda63bbcbc12d3518992da7145fa160f81085edf libraries/couchbase-lite-java-core (1.3.0-5-gcda63bb)

Related issues: #1327 #1314

vbenso:
Issue #1421 
Allow lastError to be cleared from outside once the error is handled.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -460,6 +460,12 @@ public class Replication
     protected void setLastError(Throwable lastError) {
         this.lastError = lastError;
     }
+    /**
+     * Clear lastError from outside once the error is handled
+     */
+    public void clearLastError() {
+        this.lastError = null;
+    }
 
     /**
      * Following two methods for temporary methods instead of CBL_ReplicatorSettings implementation.


hideki:
@vbenso Thank you for your PR.
We will review as soon as we released 1.3.1.

hideki:
`clearLastError()` is not public API. With current implementation of CBL Android/Java, this method is required. We should revisit this in future.

pasin:
I think we shouldn't merge this feature as it requires new API changes. If this feature is required, it should be listed as cross-platform and created a new proposed API at couchbase-lite-api repo.

hideki:
@vbenso,
After I discussed with team mate, clear lastError variable by application might cause the problem. So I reverted this commit. We will treat the issue as bug, and we will fix this as soon as possible.

vbenso:
@hideki 
Ok, 
I'll handle this in another way. 
Thanks for the effort. 
:+1:

PR #1461, opened 2016-10-03
Fixed Android Issue 1009 - attempt to re-open an already-closed object

hideki:
Ticket: couchbase/couchbase-lite-android#1009

Issue: Database class implementation allows to close database while other thread is doing read/write database operation.

Solution: By implementing reference count for store instance, `Database.close()` method waits till reference count becomes `0`. In addition,  `Database` class's methods throws RuntimeException if database is already closed.

- `CouchbaseLiteRuntimeException`: If methods throws `CouchbaseLiteException`, it requires API change, and user needs to rewrite (adding try-catch block). If `CouchbaseLiteException ` extends from `RuntimeExcepiton`, user will not be forced to catch `CouchbaseLiteException`. So I decided to add new Exception which is extended from `RuntimeException`.
- `RefCounter`: just managing number of reference. This is thread-safe.
- `Database` class: All methods which access to internal database check if database is open, and retain reference count, at end of methods release reference.
- DatbaseUpgrade, Document, View, PullerInternal, and Router classes: Just modified accordingly.

### DIFF
--- a/src/main/java/com/couchbase/lite/Database.java
+++ b/src/main/java/com/couchbase/lite/Database.java
@@ -102,7 +102,6 @@ public class Database implements StoreDelegate {
     private String name;
 
     private final AtomicBoolean open = new AtomicBoolean(false);
-    private final AtomicBoolean opening = new AtomicBoolean(false);
     private final AtomicBoolean closing = new AtomicBoolean(false);
     private final RefCounter storeRef = new RefCounter();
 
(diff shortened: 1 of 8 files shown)


pasin:
This is what I'm thinking:
1. Reference counting is managed by database and store.
2. Database provides a method that accepts a runnable kind of class to manage reference counting of multiple database operations as a whole.
3. If the code outside Database is already run in Transaction, the reference counting is already managed.
4. If the code outside Database is doing a single database operation, the reference counting is already managed.
5. If the code outside Database is doing multiple database operations, wrapping it with # or in transaction.
6. Revise the code outside database that calls Store directly to see if it makes sense to move the logic to Database class.

What do you think?

hideki:
@pasin Applied your feedbacks to PR. please re-review this? Thanks!

src/main/java/com/couchbase/lite/Database.java
```diff
@@ -1966,7 +2143,7 @@ public long getStartTime() {
      */
     @InterfaceAudience.Private
     public boolean isOpen() {
-        return open.get();
+        return (open.get() && !closing.get()) || opening.get();
```
pasin: If opening is true, doesn't it mean it's still opening but not done opening yet?

hideki: I removed opening variable from Database class. it is no longer necessary.

hideki:
@pasin Please re-review PR? Thanks!

PR #1462, opened 2016-10-05
Fixed #878 - Fixed canonicalJSON output.

### LINKED ISSUE #878: RevID generated from different API levels may not be the same
Unit Test: https://github.com/couchbase/couchbase-lite-android/blob/master/src/androidTest/java/com/couchbase/lite/RevisionsTest.java#L369-L396

Currently the test fails when running on API 19 ARM emulator.

hideki:
- to make the natural ordering of json keys, applied SortedMap/TreeMap recursively before calling `ObjectMapper.writeValueAsBytes(...)`

### DIFF
--- a/src/main/java/com/couchbase/lite/support/RevisionUtils.java
+++ b/src/main/java/com/couchbase/lite/support/RevisionUtils.java
@@ -29,6 +29,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 
 /**
  * Created by hideki on 6/17/15.
@@ -132,11 +134,10 @@ public class RevisionUtils {
      */
     @InterfaceAudience.Private
     public static byte[] asCanonicalJSON(Map<String, Object> props) {
-        if (props == null) {
+        if (props == null)
             return null;
-        }
 
-        Map<String, Object> properties = new HashMap<String, Object>(props.size());
+        Map<String, Object> properties = new TreeMap<String, Object>();
         for (String key : props.keySet()) {
             boolean shouldAdd = false;
             if (!key.startsWith("_")) {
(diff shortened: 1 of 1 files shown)


pasin:
@hideki I merged the PR but I think there is a bug. Basically the PR doesn't handle the case of an array containing map. I'm going to revert the merge.

src/main/java/com/couchbase/lite/support/RevisionUtils.java
```diff
@@ -163,6 +168,20 @@ public static String parseRevIDSuffix(String rev) {
         return json;
     }
 
+    protected static SortedMap<String, Object> convert(Map<String, Object> map) {
+        SortedMap<String, Object> dest = new TreeMap<String, Object>();
+        for (Map.Entry<String, Object> entry : map.entrySet()) {
+            String key = entry.getKey();
+            Object value = entry.getValue();
+            if (value instanceof Map) {
```
pasin: Need to handle array cases as the array can contain map.

src/main/java/com/couchbase/lite/support/RevisionUtils.java
```diff
@@ -150,7 +151,11 @@ public static String parseRevIDSuffix(String rev) {
                 return null;
             }
             if (shouldAdd) {
-                properties.put(key, props.get(key));
+                Object value = props.get(key);
+                if (value instanceof Map)
+                    properties.put(key, convert((Map) value));
+                else
```
pasin: Need to handle array cases as the array can contain map.

hideki:
@pasin 

> 1. If a map is already TreeMap, we don't need to recreate a new map.

Nested map might not be TreeMap. So still need to iterate all branches.

I agree with all other comments.

pasin:
> Nested map might not be TreeMap. So still need to iterate all branches.

Yes, but there is no need to create a new TreeMap and copy all of the items that are not Non-Sorted Map over.

pasin:
I just realized on thing though. If the Map we are discussing here is the exact same object that user set or using, we may not want to alter the Map. Copy everything seems to be more correct.

hideki:
@pasin 

> 2. If a map has only one element, we don't need to recreate a new map.

I also did not apply this suggestion. As value of Map (Dictionary) could be another Map.

The code always creates new Map and List instances.

PR #1485, opened 2016-10-14
Fixed #1126 - ChangeTracker should stop for permanent error (ex 404)

### LINKED ISSUE #1126: ChangeTracker should stop for permanent error (ex 404)
https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLRestPuller.m#L268-L289

Two years ago, we made ChangeTracker not stop for permanent error for example 404. However, current expectation, it should stop.
https://github.com/couchbase/couchbase-lite-java-core/issues/241

hideki:
- Previously ChangeTracker does not check if error is permanent error which check includes 404 or 40x check.

### DIFF
--- a/src/main/java/com/couchbase/lite/util/Utils.java
+++ b/src/main/java/com/couchbase/lite/util/Utils.java
@@ -84,14 +84,6 @@ public class Utils {
         return (code >= 400 && code <= 405) || (code == 407) || (code >= 409 && code <= 499);
     }
 
-    public static boolean isTransientErrorWithSpecialCases(Response response) {
-        return isTransientErrorWithSpecialCases(response.code());
-    }
-
-    public static boolean isTransientErrorWithSpecialCases(int code) {
-        return code == 406 || code == 408 || isTransientError(code);
-    }
-
     /**
      * in CBLMisc.m
      * BOOL CBLMayBeTransientError( NSError* error )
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/util/Utils.java
```diff
@@ -81,6 +84,14 @@ public static boolean isPermanentError(int code) {
         return (code >= 400 && code <= 405) || (code == 407) || (code >= 409 && code <= 499);
     }
 
+    public static boolean isTransientErrorWithSpecialCases(Response response) {
+        return isTransientErrorWithSpecialCases(response.code());
+    }
+
+    public static boolean isTransientErrorWithSpecialCases(int code) {
+        return code == 406 || code == 408 || isTransientError(code);
```
pasin: Why make these special cases instead of putting them into isTransientError()?

pasin: If CouchDB could send 406 and 408, shouldn't CBL iOS checks that as well? If iOS/.NET should check it, could you please check a ticket on those platforms?

For the answer, IMO, I don't think we should have a special case method.

hideki: http://docs.couchdb.org/en/latest/api/basics.html#http-status-codes

> 406 - Not Acceptable
> The requested content type is not supported by the server.

In general, this should be permanent error.

src/main/java/com/couchbase/lite/util/Utils.java
```diff
@@ -124,11 +116,8 @@ public static boolean isTransientError(Response response) {
     }
 
     public static boolean isTransientError(int code) {
-        if (code == 500 || code == 502 || code == 503 || code == 504 ||
-                code == Status.REQUEST_TIMEOUT) {
-            return true;
-        }
-        return false;
+        // Note: Status.NOT_ACCEPTABLE = 406, Status.REQUEST_TIMEOUT = 408
+        return code == 406 || code == 408 || code == 500 || code == 502 || code == 503 || code == 504;
```
pasin: In general a transient error means that it's temporary and recoverable. For 406 with CouchDB, is it recoverable?

pasin: I started to think that I don't understand the issue.

hideki:
@pasin Please re-review. Thanks!

PR #1486, opened 2016-10-15
Round progress to floor instead of rounding up to prevent misleading progress = 100

swaheed2:
When the app is syncing > 30K documents on Android, the `changes` is so close to the `Processed`:

``` 
status": "Processed A / B changes",
progress: 100
 ```
A is so close to B that progress is always 100 which is misleading. 

**Example from actual sync:**
Total documents that needed to be synced: 385,000
```
"status": "Processed 68108 / 68308 changes",
"progress": 100,
```
Both Processed and changes stay very close together.

Can progress be rounded down?


hideki:
Hi @swaheed2,
Thank you very much for your contributions!

PR #1505, opened 2016-10-28
Fixed 1023 - Runtime exception while migrating db from 1.0.0 -> 1.3.1

hideki:
Issue: https://github.com/couchbase/couchbase-lite-android/issues/1023

Problem & Solution:
- 1.0.x's path to attachment is `<db>/attachments`, 1.1.x/1.2.x 's path is `<db> attachments`. DatabaseUpgrade codes does not care v1.0.x attachment path.
- Plus more logging for error cases.

### DIFF
--- a/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
+++ b/src/main/java/com/couchbase/lite/DatabaseUpgrade.java
@@ -99,8 +99,10 @@ public final class DatabaseUpgrade {
             }
 
             // Move attachment storage directory:
-            if (!moveAttachmentsDir())
+            if (!moveAttachmentsDir()) {
+                Log.e(TAG, "Upgrade failed: Couldn't move attachment store. db: %s", db.toString());
                 return false;
+            }
 
             // Upgrade documents:
             // CREATE TABLE docs (doc_id INTEGER PRIMARY KEY, docid TEXT UNIQUE NOT NULL);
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/DatabaseUpgrade.java
```diff
@@ -167,11 +173,16 @@ protected void deleteSQLiteFiles() {
...
     private boolean moveAttachmentsDir() {
+        // v1.1.x or v1.2.x
         File oldAttachmentsPath = new File(FileDirUtils.getPathWithoutExt(path) + " attachments");
-        if (!oldAttachmentsPath.canRead())
-            return true;
+        if (!oldAttachmentsPath.exists() || !oldAttachmentsPath.canRead()) {
+            // v1.0.x
+            oldAttachmentsPath = new File(FileDirUtils.getPathWithoutExt(path), "attachments");
+            if (!oldAttachmentsPath.exists() || !oldAttachmentsPath.canRead())
+                return false;
```
pasin: Even though i merged the PR, can u confirm that return fase is correct. I don't know why previously return true.

hideki: In case of not existing the attachment directory, the method should return true, then go through. Thank you very much for catching this.

PR #1508, opened 2016-10-31
Fixed #1507 - Inappropriate Exception message is thrown from SQLite

### LINKED ISSUE #1507: Inappropriate Exception message is thrown from SQLite when database corruption happens
https://github.com/couchbase/couchbase-lite-android/issues/985#issuecomment-257302824

hideki:
Inappropriate Exception message is thrown from SQLite when database corruption happens.
https://github.com/couchbase/couchbase-lite-java-core/issues/1507

### DIFF
--- a/src/main/java/com/couchbase/lite/storage/SQLiteStorageEngineBase.java
+++ b/src/main/java/com/couchbase/lite/storage/SQLiteStorageEngineBase.java
@@ -54,7 +54,7 @@ public abstract class SQLiteStorageEngineBase implements SQLiteStorageEngine {
 
     @Override
     public boolean open(String path, SymmetricKey encryptionKey) throws SQLException {
-        if(database != null && database.isOpen())
+        if (database != null && database.isOpen())
             return true;
 
         boolean hasError = false;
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/storage/SQLiteStorageEngineBase.java
```diff
@@ -72,12 +72,20 @@ public boolean open(String path, SymmetricKey encryptionKey) throws SQLException
...
             Log.v(Log.TAG_DATABASE, "%s: Opened Android sqlite db", this);
-        } catch(SQLiteDatabaseCorruptException e) {
+        } catch (SQLiteDatabaseCorruptException e) {
             hasError = true;
-            Log.e(Log.TAG_DATABASE, "Unauthorized to open the SQLite database", e);
-            throw new SQLException(SQLException.SQLITE_ENCRYPTION_UNAUTHORIZED,
-                    "Cannot decrypt or access the database");
-        } catch(com.couchbase.lite.internal.database.SQLException e) {
+            // SQLCipher with encryption
+            if (encryptionKey != null && supportEncryption()) {
```
pasin: This check doesn't cover 2 things:
1. If database is encrypted but the library used is sqlite instead of sqlcipher.
2. If database is encrypted but no encryption key given.

pasin:
Database encryption error and database corruption error are sharing the same error code.  I don't see a good way to differentiate between encryption error and database corruption. But database corruption is not expected to happen anyway.

hideki:
@pasin please review again.

pasin:
How's about not fixing this?

hideki:
Currently some users are observing "Unauthorized to open the SQLite database" error message which was caused by real database file corruption. In case user does not activate SQLCipher, this error message confuses us, and mislead us to investigate different area.

pasin:
> this error message confuses us, and mislead us to investigate different area.

If this is the case, I don't see the new message is better than the original one because I still don't know if users use SQLite or SQLCipher or not.

pasin:
Also even if users use SQLCipher, there is no way to differentiate between Encryption error or Corruption. So I feel like fixing this doesn't help investigating the issue much comparing to not fixing it.

hideki:
This way of fix makes identify if user inputed correct password in case SQLCipher is enabled. We might need to find better solution.

PR #1515, opened 2016-11-12
Fixed #1481 - ConnectionError: ('Connection aborted.', error(54, 'Con…

### LINKED ISSUE #1481: ConnectionError: ('Connection aborted.', error(54, 'Connection reset by peer'))
This is using 1.3.1-61

When running certain tests, I hit this intermittently with the Android listener. 

Here is an example of one such failure - http://uberjenkins.sc.couchbase.com/job/net-listener-client-client/7/testReport/testsuites.listener.shared.client_client/test_peer_2_peer_sanity/test_peer_2_peer_sanity_pull/

We need to determine root cause, I have talked with @hideki and we think it may be due to the thread limit on the Android listener. 

I am able to reproduce so we should pair debug this.

hideki:
…nection reset by peer'))

Connection abort is actually caused by out-of-memory. And OOM is caused by requests are queued up for push replication, as client send POST or PUT /{db}/{doc} too fast. As solution, I made the logic of adding revisions to push replicator from listener thread to synchronous mode instead of asynchronous.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PusherInternal.java
@@ -295,12 +295,25 @@ public class PusherInternal extends ReplicationInternal implements Database.Chan
     @InterfaceAudience.Private
     public void changed(Database.ChangeEvent event) {
         final List<DocumentChange> changes = event.getChanges();
-        supportExecutor.submit(new Runnable() {
-            @Override
-            public void run() {
-                submitRevisions(changes);
-            }
-        });
+
+        // NOTE: REST API client can send POST/PUT /{db}/{doc} request too faster than
+        // push replicator can handle. If request comes from listener thread, submitting revision
+        // to batcher synchronusly. Otherwise, asynchronously.
+        // ref: https://github.com/couchbase/couchbase-lite-java-core/issues/1481
+        if (calledFromListener()) {
+            submitRevisions(changes);
+        } else {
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/PusherInternal.java
```diff
@@ -295,12 +295,25 @@ public void beginReplicating() {
...
+        } else {
+            supportExecutor.submit(new Runnable() {
+                @Override
+                public void run() {
+                    submitRevisions(changes);
+                }
+            });
+        }
+    }
+
```
pasin: I have a couple of questions.
 1. If I remember correctly, the submitRevisions() is called in the supportExecutor to fix an issue, probably related to threading issues. Is it ok to call submitRevisions outside the supportExecutor?
 2. From the note above that REST API client can send POST/PUT /{db}/{doc} request too faster than push replicator can handle, isn't the pause logic in push replicator supposed to prevent this issue? Also Can't the native android application insert the documents as fast as the REST client?

hideki: 1. I added supportExecutor to avoid that caller is blocked by pause logic.
2. pause logic itself works fine. However, change events come faster, so changes and associated RevisionInternal is stuck in memory.
3. Native API could cause this issue too. 

Currently I don't have clean way to make REST client slow down the requests.

We could discuss this on Monday.

PR #1521, opened 2016-11-28
Fixed #1495 - Changes feed skipping changes on rapid long poll

### LINKED ISSUE #1495: Changes feed skipping changes on rapid long poll
Using 1.3.1-60, I ran @sethrosetter 's test_peer_2_peer_sanity_pull (using SQLCipher for both) with .NET as the receiver and Android as the sender (i.e. pull from Android to .NET) and discovered that the client often misses documents.  I opened the DB on the .NET side and found that the first missing document was test_ls_db2_160.  I also found that this document never showed up in the changes feed (among others later).  Here is the traffic around the time in question:

```
1) 10:46:00.454409: PUT /ls_db2/test_ls_db2_159 (from test framework)
2) 10:46:00.455935: POST /ls_db2/_changes (from LiteServ .NET, since 159)
3) 10:46:00.458387: 200 response to (2) contains test_ls_db2_159 (last_seq=160)
4) 10:46:00.458396: 201 response to (1)
5) 10:46:00.460067: PUT /ls_db2/test_ls_db2_160 (from test framework)
6) 10:46:00.461837: POST /ls_db2/_all_docs (from LiteServ .NET)
7) 10:46:00.461868: POST /ls_db2/_changes (from LiteServ .NET since 160)
8) 10:46:00.468612: 201 response to (5)
9) 10:46:00.470141: PUT /ls_db2/test_ls_db2_161 (from test framework)
10) 10:46:00.475547: 200 response to (7) *contains test_ls_db_161 only* (last_seq=162)
11) 10:46:00.475556: 201 response to (9)
12) 10:46:00.477242: 200 response to (6) (contains doc data for 106-159)
```

hideki:
- About issue #1495, In race condition, there is possibility database change event is fired between `changesSince()` method and starting to observe DatabaseChange event. For this issue, CBL Java takes solution from .NET implementation. In the first DatabaseChange event, calling `changesSince()` method.
- make consistent since variable/parameter should be long (not int)
- Added some null check
- Removed `sendLongpollChanges(List<RevisionInternal> revs)` method which allows to send response since value is `0`.
- In case of sending response from DatabaseChange event, it ignores `ChangeOptions.isIncludeConflict` parameter.
- In case of sending response from DatabaseChange event with feed=long but revs.size() is 0, should start timeout. Otherwise, I am not sure how _changes API behave.

### DIFF
--- a/src/main/java/com/couchbase/lite/router/Router.java
+++ b/src/main/java/com/couchbase/lite/router/Router.java
@@ -113,6 +113,9 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
     private URL source = null;
 
     // _changes request:
+    private long changesSince = 0;
+    private boolean filled = false;
+    private ChangesOptions changesOptions = null;
     private ReplicationFilter changesFilter;
     Map<String, Object> changesFilterParams = null;
     private boolean longpoll = false;
@@ -204,6 +207,19 @@ public class Router implements Database.ChangeListener, Database.DatabaseListene
             }
         }
 
+        return result;
+    }
+
+    private long getLongQuery(String param, long defaultValue) {
+        long result = defaultValue;
+        String value = getQuery(param);
+        if (value != null) {
+            try {
+                result = Long.parseLong(value);
+            } catch (NumberFormatException e) {
+                //ignore, will return default value
+            }
+        }
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


hideki:
Note: This PR cause the test failure with `test_peer_2_peer_sanity_pull` functional test. I will close this PR. We need to rework for this issue.

Issue #1522, opened 2016-11-28
Router: Implement multi-part doc support


hideki:
Router does not support multi-part document.  Push replicator sends multi-part doc first, then fallback to JSON request. Need to support multi-part doc.

PR #1523, opened 2016-11-28
Fixed couchbase lite p2p attachment replication on android (android i…

hideki:
…ssue 1037)

original ticket: https://github.com/couchbase/couchbase-lite-android/issues/1037

There are two reasons that makes p2p attachment replication fails.
1. Router can not handle multi-part requests
2. Push replicator does not execute fall-back solution (push by JSON) because Router returns http error 406 instead of 415.

In this commit, fix case 2. Case 1 I will create new ticket.


hideki:
ticket for multi-part doc support: https://github.com/couchbase/couchbase-lite-java-core/issues/1522

PR #1538, opened 2017-01-04
Fixed #1534 - Push Replication Fails With createTarget Set

### LINKED ISSUE #1534: Push Replication Fails With createTarget Set
Client – Couchbase Lite 1.3.1
Remote – CouchDB 2.0

When doing a push replication with no target database present on the remote CouchDB instance, and after having set the `createTarget` property in the Replication object to true (via calling `setCreateTarget(true)`), Couchbase Lite fails to create the target database and the replication stops. If I look at the network traffic between the mobile and my remote database, Couchbase Lite actually does a GET instead of a PUT to create the remote database. The GET returns a 404, stopping replication.

It appears that this issue is due to code in PusherInternal.java and RemoteRequest.java. PusherInternal's `maybeCreateRemoteDB()` gets called after the GET from _local checkpoint fails, and this initiates a PUT request to create the remote DB as follows (from PusherInternal.java):

```
	protected void maybeCreateRemoteDB() {
		if (!createTarget) {
			return;
		}
		creatingTarget = true;
		Log.v(TAG, "Remote db might not exist; creating it...");

		Future future = sendAsyncRequest("PUT", "", null, new RemoteRequestCompletion() { 
              ...
```

It looks like the above `sendAsyncRequest()` call ends up creating a RemoteRequest object to fulfill the request. The call to set the request to a PUT looks like the following (from RemoteRequest.java):

```
	protected Request.Builder setBody(Request.Builder builder) {
		if (body != null) {
			byte[] bodyBytes = null;
			try {
				bodyBytes = Manager.getObjectMapper().writeValueAsBytes(body);
                      ...

			if ("PUT".equalsIgnoreCase(method))
				builder.put(requestBody);
			else if ("POST".equalsIgnoreCase(method))
				builder.post(requestBody);
		}
		return builder;
	}
```

The code that should turn the request object into a PUT request is contained in a block of code that only gets called if the body object is not null, but the body is set to null in the `sendAsyncRequest()` call, so the request goes out as a GET. To test this, I modified Couchbase Lite to insert a dummy body in the `maybeCreateRemoteDB()` code, and that got it working, but it was a gross hack.

It appears that the okhttp3 library that is being used may not allow creation of PUT requests without a body, so simply adding a `builder.put(requestBody)` call outside of the if statement might not work.

hideki:
OkHttp requires to set the body with calling `put(…)`. Otherwise the request becomes GET request.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/RemoteRequest.java
+++ b/src/main/java/com/couchbase/lite/replicator/RemoteRequest.java
@@ -53,8 +53,7 @@ public class RemoteRequest implements CancellableRunnable {
     // Don't compress data shorter than this (not worth the CPU time, plus it might not shrink)
     public static final int MIN_JSON_LENGTH_TO_COMPRESS = 100;
     public static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
-    public static final MediaType TEXT = MediaType.parse("text/plain; charset=utf-8");
-
+    
     ////////////////////////////////////////////////////////////
     // Member variables
     ////////////////////////////////////////////////////////////
--- a/src/main/java/com/couchbase/lite/replicator/RemoteRequest.java
+++ b/src/main/java/com/couchbase/lite/replicator/RemoteRequest.java
@@ -53,6 +53,7 @@ public class RemoteRequest implements CancellableRunnable {
     // Don't compress data shorter than this (not worth the CPU time, plus it might not shrink)
     public static final int MIN_JSON_LENGTH_TO_COMPRESS = 100;
     public static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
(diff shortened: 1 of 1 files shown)


src/main/java/com/couchbase/lite/replicator/RemoteRequest.java
```diff
@@ -53,6 +53,7 @@
     // Don't compress data shorter than this (not worth the CPU time, plus it might not shrink)
     public static final int MIN_JSON_LENGTH_TO_COMPRESS = 100;
     public static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
+    public static final MediaType TEXT = MediaType.parse("text/plain; charset=utf-8");
```
pasin: Is the TEXT constant used somewhere?

hideki: no. I will remove it.

PR #1583, opened 2017-02-08
Fixed #1581 - Replication failures when running Sync Gateway in distr…

### LINKED ISSUE #1581: Replication failures when running Sync Gateway in distributed index
Android - 1.4-36
SG version - 1.4-65
SG mode - DI
CBS version - 4.5.1

Test failures - http://uberjenkins.sc.couchbase.com/job/android-listener-client-sync-gateway/129/. These tests pass when running in channel cache.

According to @hideki, the replicators are all in a wait state.

hideki:
…ibuted index

Calling `Batcher.flushAll(true)` concurrently causes dead-lock. Need to avoid it.

### DIFF
--- a/src/main/java/com/couchbase/lite/support/Batcher.java
+++ b/src/main/java/com/couchbase/lite/support/Batcher.java
@@ -53,6 +53,7 @@ public class Batcher<T> {
 
     private final Object mutex = new Object();
     private final Object processMutex = new Object();
+    private final Object flushAllMutext = new Object();
 
     ///////////////////////////////////////////////////////////////////////////
     // Constructors
@@ -158,6 +159,12 @@ public class Batcher<T> {
                     processMutex.wait(5);
                 } catch (InterruptedException e) { }
             }
+        }
+    }
+
+    public void flushAllAndWait() {
+        synchronized (flushAllMutext) {
+            flushAll(true);
         }
     }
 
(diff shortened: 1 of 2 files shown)


hideki:
@pasin If you have a chance, could you please review this PR even though I already merged this when you will review?

src/main/java/com/couchbase/lite/support/Batcher.java
```diff
@@ -161,6 +162,12 @@ public void queueObjects(List<T> objects) {
         }
     }
 
+    public void flushAllAndWait() {
+        synchronized (flushAllMutext) {
+            flushAll(true);
+        }
+    }
+
```
pasin: I think it's better to fix this in the flushAll method instead of having another flush method.

Also can the batcher be flushed without waiting to finish?

hideki: > Also can the batcher be flushed without waiting to finish?

Without wait, it causes OOM with large data set (I have used for testing).

> I think it's better to fix this in the flushAll method instead of having another flush method.

sure. I will update it.

hideki: REF: https://github.com/couchbase/couchbase-lite-java-core/issues/1586

hideki: Today, 02/08/17, is just a day before FC (Feature Complete). We will work for this during next release cycle.

Issue #1403, opened 2016-08-25
usePost flag in ChangeTracker should be editable


hideki:
https://github.com/couchbase/couchbase-lite-android/issues/978

hideki:
API Request: https://github.com/couchbaselabs/couchbase-lite-api/issues/162

PR #1598, opened 2017-02-16
Fixed #978 Made usePOST modifiable

AfzalivE:
for use in PullerInternal (Filtered pull replications from Bluemix/Cloudant).

https://github.com/couchbase/couchbase-lite-java-core/issues/1403
https://github.com/couchbase/couchbase-lite-android/issues/978

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/PullerInternal.java
@@ -180,6 +180,7 @@ public class PullerInternal extends ReplicationInternal implements ChangeTracker
                 changeTracker.setFilterParams(filterParams);
             }
         }
+        changeTracker.setUsePOST(usePOST);
         changeTracker.setDocIDs(documentIDs);
         changeTracker.setRequestHeaders(requestHeaders);
         changeTracker.setContinuous(lifecycle == Replication.Lifecycle.CONTINUOUS);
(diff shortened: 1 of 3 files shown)


hideki:
@AfzalivE 

We have been discussed about the issue of `POST /_changes` with Cloudant internally. It seems Cloudant can not handle `POST /_changes` request. We consider this is Cloudant issue.

To be consistent with other platforms, CBL iOS and .NET,  we can not merge your PR into master branch just for Android/Java platform. Can you use your fix with your branch?

Thanks!
Hideki

AfzalivE:
Yes, I've been using my own version for now since 1.3.1 works quite well for us. 

Hmm interesting, that makes sense. I don't have a lot of experience with iOS or .NET but if I make a change like this for those as well, would that help?

hideki:
@AfzalivE we discussed about adding `setUsePOST()` method. At that moment, we decided that we don't implement it. We hope Cloudant fixes this problem as CouchDB does not have this problem.

Also, I understand Cloudant users would like to use GET because of Cloudant pricing model. https://cloudant.com/product/pricing/

AfzalivE:
Fair. Would be great if they do fix it, but yesterday, the POST request was actually timing out so probably some bug as well that they could probably fix.

No worries. Great project btw, syncing is spot on!

hideki:
Hi @AfzalivE,
Thank you very much again for your contribution.
However, we can not merge this PR. I will close this.
Thanks again!
Hideki

PR #1600, opened 2017-02-19
logging improvements

npomfret:


npomfret:
FYI - the reason I've made this change is that my CBL instance has an error which is stopping replication from working.  There's no stack trace to be found though.  After some debugging it appears that there's plenty of places in the code base where useful exception stacks are being thrown away.  I've done my best to find them all in this PR.

hideki:
Hi @npomfret,

I am fine to merge commit #e4779ed. But I am not sure about commit #4041b2d. If all errors have a printed stacktrace in the logcat, it is convenient for debugging. But some users worry about any Exceptions in the logcat. We have tried to not print StackTrace for all errors, only selected ones. But no clear guideline to print or not to print.

npomfret:
I understand but I would have thought the log level was the important thing
here. An exception logged at error is something to worry about, and one at
warn, well less so.

Without exceptions being logged you have little chance of diagnosing
unexpected errors. Which is the situation I'm in. Cbl is crashing and my
app doesn't work - but no stack, no clues - it's just dead.

Perhaps you could keep some changes, the log.error ones, and increase the
log level of the more serious log.warn calls? I don't know the code base so
am unsure of what is really important and is possibly logged at the wrong
level.

npomfret:
... or switch the default log level to error?

npomfret:
Or if that isn't acceptable, how about double logging?  For each exception you _don't_ want to see, we log it at VERBOSE level, and immediately after we log the warning message without the stack trace .

hideki:
@npomfret 
Thank you very much for your contribution.
However, we can not merge this.
We will review our logging with 2.0 release time frame.
Thanks!

PR #1618, opened 2017-04-13
Fixed #1617 - The specified SyncGatewaySession value with /_replicate REST API is not used.

### LINKED ISSUE #1617: The specified SyncGatewaySession value with /_replicate REST API is not used.
REF: CBSE-3606

Version: Couchbase Lite Android/Java v1.4.0 or earlier

### The root cause of newly set SyncGatewaySesion value is not used.

1. CookieJar and how it is used:
Cookie values are stored with only responses from REST API call. It means the Cookie is stored if HTTP response header contains `SetCookie:` header value. In other scenarios, CBL does not store Cookie in the CookieJar.

2. What logic uses cookies in CookieJar
OkHttp always loads cookie values from CookieJar before sends a request to the server. Following is stacktrace to prove it:
```
 at com.couchbase.lite.support.PersistentCookieJar.loadForRequest(PersistentCookieJar.java:157)
 at okhttp3.internal.http.HttpEngine.networkRequest(HttpEngine.java:566)
 at okhttp3.internal.http.HttpEngine.sendRequest(HttpEngine.java:199)
 at okhttp3.RealCall.getResponse(RealCall.java:243)
 at okhttp3.RealCall$ApplicationInterceptorChain.proceed(RealCall.java:201)
 at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:163)
 at okhttp3.RealCall.execute(RealCall.java:57)
 at com.couchbase.lite.replicator.RemoteRequest.executeRequest(RemoteRequest.java:261)
 at com.couchbase.lite.replicator.RemoteRequest.execute(RemoteRequest.java:165)
 at com.couchbase.lite.replicator.RemoteRequest.run(RemoteRequest.java:105)
 at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:428)
 at java.util.concurrent.FutureTask.run(FutureTask.java:237)
 at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:272)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
 at java.lang.Thread.run(Thread.java:761)
```
3. Based on above two info, till cookie value is stored in the CookieJar, the replicator uses SyncGatewaySession value, which is set by _replicator REST API, in the memory.
4. What makes Cookies are stored into the CookieJar.
SyncGateway responds with `SetCookie:` header value in case SyncGatewaySession will be expired in 10% of total TTL. This is SyncGatewaySession auto-update feature. So this makes Cookie to be stored into the CookieJar.
5. Once a cookie is stored in CookieJar, this cookie is used till cookie expire or overwritten. Also, newly set Cookie (SyncGatewaySession) is overwritten by a cookie in CookieJar just before sending Http Request.

### Workaround
- When /_replicate REST API is called with `Cookie: SyncGatewaySession=....` header value, CBL stores it into CookieJar. So OkHttp always uses newly set SyncGatewaySession value.

hideki:
Please refer https://github.com/couchbase/couchbase-lite-java-core/issues/1617#issue-220851080 for details.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -921,8 +921,7 @@ public class Replication
     public void setHeaders(Map<String, Object> requestHeadersParam) {
         properties.put(ReplicationField.REQUEST_HEADERS, requestHeadersParam);
         replicationInternal.setHeaders(requestHeadersParam);
-
-        // NOTE:
+        // Fix - https://github.com/couchbase/couchbase-lite-java-core/issues/1617
         storeSyncGatewaySesionIntoCookieJar(requestHeadersParam);
     }
 
@@ -932,6 +931,8 @@ public class Replication
                 String cookieString = (String) requestHeadersParam.get("Cookie");
                 if (cookieString.indexOf("SyncGatewaySession") != -1) {
                     if (remote != null) {
+                        // NOTE: In case that the path of URL is not end with `/`,
+                        //       last segment of a path is not stored as a path of Cookie.
                         Cookie cookie = Cookie.parse(HttpUrl.get(remote), cookieString);
                         if (cookie != null && replicationInternal != null) {
                             replicationInternal.setCookie(cookie);
--- a/src/main/java/com/couchbase/lite/replicator/Replication.java
+++ b/src/main/java/com/couchbase/lite/replicator/Replication.java
@@ -922,21 +922,19 @@ public class Replication
         properties.put(ReplicationField.REQUEST_HEADERS, requestHeadersParam);
         replicationInternal.setHeaders(requestHeadersParam);
         // Fix - https://github.com/couchbase/couchbase-lite-java-core/issues/1617
-        storeSyncGatewaySesionIntoCookieJar(requestHeadersParam);
+        storeCookiesIntoCookieJar(requestHeadersParam);
     }
 
-    private void storeSyncGatewaySesionIntoCookieJar(Map<String, Object> requestHeadersParam) {
+    private void storeCookiesIntoCookieJar(Map<String, Object> requestHeadersParam) {
         try {
             if (requestHeadersParam.containsKey("Cookie") && requestHeadersParam.get("Cookie") instanceof String) {
                 String cookieString = (String) requestHeadersParam.get("Cookie");
-                if (cookieString.indexOf("SyncGatewaySession") != -1) {
-                    if (remote != null) {
-                        // NOTE: In case that the path of URL is not end with `/`,
-                        //       last segment of a path is not stored as a path of Cookie.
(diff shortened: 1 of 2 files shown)


src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -918,6 +921,26 @@ public String getFilter() {
...
+
+    private void storeCookiesIntoCookieJar(Map<String, Object> requestHeadersParam) {
+        try {
+            if (requestHeadersParam.containsKey("Cookie") && requestHeadersParam.get("Cookie") instanceof String) {
+                String cookieString = (String) requestHeadersParam.get("Cookie");
+                if (remote != null) {
+                    // NOTE: In case that the path of URL is not end with `/`,
+                    //       last segment of a path is not stored as a path of Cookie.
+                    Cookie cookie = Cookie.parse(HttpUrl.get(remote), cookieString);
+                    if (cookie != null && replicationInternal != null) {
```
pasin: I think we don't need to check if the replicationInternal is null here as it will not be null?

pasin: Does it also need to remove the Cookies from the requestHeadersParams as well? I don't know the behavior when the cookies are in both headers and cookie jar.

hideki: ### for comment 1
- replicationInternal should not be null. So I will remove this.

### for comment 2
- Cookie's in Cookie jar always overwrite if it exists. However, passed cookie value in the headers does not includes expirers, domain, and path. So it could send two cookies in the final request header. As you advise, I feel it is better to remove from header.

hideki:
@pasin applied feedback from you. Please review this PR again?

src/main/java/com/couchbase/lite/replicator/Replication.java
```diff
@@ -916,10 +919,33 @@ public String getFilter() {
...
+        // Fix - https://github.com/couchbase/couchbase-lite-java-core/issues/1617
+        storeCookiesIntoCookieJar(requestHeadersParam);
+
         properties.put(ReplicationField.REQUEST_HEADERS, requestHeadersParam);
         replicationInternal.setHeaders(requestHeadersParam);
     }
 
+    private void storeCookiesIntoCookieJar(Map<String, Object> requestHeadersParam) {
+        try {
+            if (requestHeadersParam.containsKey("Cookie") && requestHeadersParam.get("Cookie") instanceof String) {
```
pasin: I think requestHeadersParam can be null so null check would be required.

hideki: thanks!

hideki:
@pasin can you re-review?

PR #1638, opened 2017-06-21
Fixed #1637 - 1.x: RejectedExecutionException is through from Replicator

### LINKED ISSUE #1637: 1.x: RejectedExecutionException is through from Replicator
The following line of codes allows to shutdown the executor service during submitting the task to Executor.
```java
  if (executor != null && !executor.isShutdown())
    executor.shutdown();
```
https://github.com/couchbase/couchbase-lite-java-core/blob/master/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java#L1478-L1479

Links:
- https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html
- https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html
- http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/java/util/concurrent/ScheduledThreadPoolExecutor.java#ScheduledThreadPoolExecutor.shutdown%28%29

hideki:
- `Executor.submit(...)` throws `RejectedExecutionException` if `Executor` is shutdown. So we had two choices. One is apply `try-catch` block for all Executor.submit(...) method. Another one is make Executor.submit(...), schedule(...), shutdown(), etc methods are thread safe. I picked second choice because already CBL codes has tried to be thread-safe with Executor services.

### DIFF
--- a/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
+++ b/src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
@@ -229,20 +229,22 @@ abstract class ReplicationInternal implements BlockingQueueListener {
     protected void fireTrigger(final ReplicationTrigger trigger) {
         Log.d(Log.TAG_SYNC, "%s [fireTrigger()] => " + trigger, this);
         // All state machine triggers need to happen on the replicator thread
-        synchronized (executor) {
-            if (!executor.isShutdown()) {
-                executor.submit(new Runnable() {
-                    @Override
-                    public void run() {
-                        try {
-                            Log.d(Log.TAG_SYNC, "firing trigger: %s", trigger);
-                            stateMachine.fire(trigger);
-                        } catch (Exception e) {
-                            Log.i(Log.TAG_SYNC, "Error in StateMachine.fire(trigger): %s", e.getMessage());
-                            throw new RuntimeException(e);
+        if (executor != null) {
+            synchronized (executor) {
+                if (!executor.isShutdown()) {
+                    executor.submit(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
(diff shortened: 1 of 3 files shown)


hideki:
@pasin please re-review this PR? thanks!

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -229,20 +228,22 @@ public void triggerGoOnline() {
...
-                executor.submit(new Runnable() {
-                    @Override
-                    public void run() {
-                        try {
-                            Log.d(Log.TAG_SYNC, "firing trigger: %s", trigger);
-                            stateMachine.fire(trigger);
-                        } catch (Exception e) {
-                            Log.i(Log.TAG_SYNC, "Error in StateMachine.fire(trigger): %s", e.getMessage());
-                            throw new RuntimeException(e);
+        if (executor != null) {
```
pasin: Need need to check null, right?

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
```diff
@@ -190,7 +190,6 @@
     protected void finalize() throws Throwable {
         if (executor != null && !executor.isShutdown()) {
```
pasin: no need to check null here?

hideki: `finalize()` method is called even though constructor fails. the executor is instantiated in the constructor.

src/main/java/com/couchbase/lite/replicator/ReplicationInternal.java
hideki: I will fix this.

hideki:
please review this again. Thanks!

PR #1666, opened 2017-10-04
Fixed Android Issue 1229 - 1.x: CBL 1.4 and 1.3.1, replication of 200k docs : linear consumption of memory

hideki:
https://github.com/couchbase/couchbase-lite-android/issues/1229

- `ReplicationInternal.runnables` variable holds all `RemoteBulkdDownloaderRequest` instances till the end of oneshot replication. 
This PR makes `RemoteBulkdDownloaderRequest` instances will be removed from `runnables` immediately after its task is completed.


src/main/java/com/couchbase/lite/replicator/PullerInternal.java
```diff
@@ -379,6 +378,8 @@ public void onCompletion(Response httpResponse, Object result, Throwable e) {
                             --httpConnectionCount;
                             // Start another task if there are still revisions waiting to be pulled:
                             pullRemoteRevisions();
+
```
pasin: Can you remove the cancellable in here so you don't need the CancellableManager and the reverseCancellables map? 

For example:
```Java
cancellables.values().remove(this);
```

pasin: I don't have a big picture about threading (what thread will this onCompletion get called). Hope there is no race issue here.

